PR #5, opened 2021-01-08
add a workflow to manually trigger a rebase of a PR opened by ipldbot

marten-seemann:
This is useful in repositories where the update of a workflow initially
fails, and the problem is subsequently fixed. It's necessary to rebase
the PR updating the workflow to make sure that all tests now actually
pass.

The new workflow (which will distribited to every participating repo)
achieves this by running on every issue comment, and triggers a rebase
if the comment was made on a PR opened by ipldbot, and it contains the
keywords "@ipldbot rebase".

### DIFF
--- a/.github/workflows/copy-workflow.yml
+++ b/.github/workflows/copy-workflow.yml
@@ -21,7 +21,7 @@ jobs:
       fail-fast: false
       matrix:
         cfg: ${{ fromJson(needs.matrix.outputs.targets) }}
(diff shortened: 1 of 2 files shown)


rvagg (approved):
neato, but I'd like to see it in action, can we get this into the btc repo and put it in a state where we can see it work?

marten-seemann:
I'll merge this one and you can see it in action then.

PR #8, opened 2021-02-18
drop Go 1.14, add Go 1.16

marten-seemann:
cc @mvdan


marten-seemann:
Merging this PR will trigger PRs in all participating repos, updating their `go-test` workflow.

mvdan (approved):
Beware that 1.16 enables module mode by default on all cases, so this might break some very old setups if they haven't switched to modules yet.

marten-seemann:
Good to know. I see that as a feature, not a bug. We shouldn't have any repos that don't use modules.

PR #10, opened 2021-02-24
add a workflow to automatically merge ipldbot's PRs

marten-seemann:
This workflow aims to reduce the overhead of changing on our workflows. For example, we don't want to manually merge ipldbot's PRs that update the Go version every 6 months across all repos.

The `automerge` workflow runs in every participating repo and automatically merges any PR authored by ipldbot, as soon as (and only if) all checks succeed.

### DIFF
--- a/workflow-templates/automerge.yml
+++ b/workflow-templates/automerge.yml
@@ -1,0 +1,23 @@
+# Automatically merge pull requests opened by ipldbot, as soon as (and only if) all tests pass.
+# This reduces the friction associated with updating with our workflows.
+
+on: [ pull_request ]
+
+jobs:
+  automerge:
+    if: github.event.pull_request.user.login == 'ipldbot'
+    runs-on: ubuntu-latest
+    steps:
+    - name: Wait on tests
+      uses: lewagon/wait-on-check-action@v0.2
+      with:
+        ref: ${{ github.event.pull_request.head.sha }}
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        wait-interval: 10
+        running-workflow-name: 'automerge' # the name of this job
+    - name: Merge PR
+      uses: pascalgn/automerge-action@v0.13.1
+      env:
+        GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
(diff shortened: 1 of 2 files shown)


workflow-templates/automerge.yml
```diff
@@ -0,0 +1,23 @@
...
+
+on: [ pull_request ]
+
+jobs:
+  automerge:
+    if: github.event.pull_request.user.login == 'ipldbot'
+    runs-on: ubuntu-latest
+    steps:
+    - name: Wait on tests
+      uses: lewagon/wait-on-check-action@v0.2
```
Stebalien, line 12: We need to fork these actions into our org. Using externally defined actions is insecure.

Stebalien: Insecure == the author can re-tag and then get a "token" that will let them do pretty much anything.

mvdan:
> For example, we don't want to manually merge ipldbot's PRs that update the Go version every 6 months across all repos.

I think automatically adding new versions if green makes sense. I'm a bit more worried about automatically removing older versions. Perhaps a project explicitly wants to support Go 1.14.x for an extra few weeks/months until the next release, even after Go 1.16.x is out.

Maybe I'm overthinking that, though. Perhaps there would be a way to opt out of (or into) the ipldbot auto-merging.

Stebalien:
> Perhaps a project explicitly wants to support Go 1.14.x for an extra few weeks/months until the next release, even after Go 1.16.x is out.

Ideally, there would be a way to "customize" these. Can we set per-repo variables?

Stebalien:
(later, we don't need to handle that now)

marten-seemann:
@Stebalien I updated the PR to use commit hashes. Haven't done an in-depth review for those versions though.

> Ideally, there would be a way to "customize" these. Can we set per-repo variables?

Not yet, but it's planned in some form. There's the idea to use a templating engine (CUE? YAML templates?) so that repositories can add steps / configure the run, if they want to. We need to find something that strikes the right balance between simplicity and complexity.

Stebalien (approved):
Future note: we might want to add some kind of script to resolve tags to commits, update, etc. But we can do that later.

PR #11, opened 2021-02-24
add a license

marten-seemann:
Let's all be good citizens of the internet and put a license on our code, shall we?


willscott:
(for this as template should we have both mit and apache2 licenses?)

willscott (approved):

mvdan:
Yeah, I think the two license files is the standard, though @mikeal had thoughts on that :)

marten-seemann:
Makes sense. I added an Apache license.

PR #18, opened 2021-02-26
use commit hash to include 3rd party GitHub Actions in autorebase

marten-seemann:
Sorry, looks like I forgot that one in #15. 

Looks like we need a GitHub Action to check our GitHub Actions ;) (just kidding)


mvdan (approved):
I think some form of automated checks aren't entirely out of the question, if GitHub does not support that kind of restriction themselves.

marten-seemann:
Actually, this shouldn't be too hard. A few lines of `grep`, `awk` and `cut` should do the trick.

Stebalien (approved):

PR #20, opened 2021-02-28
add a workflow to check that 3rd party actions are included by hash

marten-seemann:
As suggested by @mvdan in https://github.com/ipld/.github/pull/18#pullrequestreview-599396076.

Test failing because the copy workflow hasn't been updated yet (see #21).

### DIFF
--- a/.github/workflows/check-3rd-party.sh
+++ b/.github/workflows/check-3rd-party.sh
@@ -7,7 +7,7 @@ for line in `sed -ne 's/[[:space:]-]*uses:[[:space:]]*//p' $1 | sed -e 's/\s*#.*
   author=`echo $line | awk -F/ '{print $1}'`
   if [[ $author == "actions" ]]; then continue; fi
   version=`echo $line | awk -F@ '{print $2}' | awk '{print $1}'`
-  if [[ ${#version} != 40 ]]; then
+  if ! [[ "$version" =~ ^[a-f0-9]{40}$ ]]; then
     status=1
     echo "$FILE includes $line and doesn't use commit hash"
   fi
--- a/.github/workflows/check-3rd-party.sh
+++ b/.github/workflows/check-3rd-party.sh
@@ -1,0 +1,17 @@
+#!/bin/bash
+
+FILE=$1
+status=0
+
+while read -r line; do
+  src=`echo $line | sed -E "s/(.*)uses://g" | xargs`
+  author=`echo $src | awk -F/ '{print $1}'`
+  if [ $author == "actions" ]; then continue; fi
+  version=`echo $src | awk -F@ '{print $2}' | awk '{print $1}'`
+  if [ ${#version} != 40 ]; then
+    status=1
+    echo "$FILE includes $src and doesn't use commit hash"
+  fi
+done < <(grep "uses:" $FILE)
+
+exit $status
--- a/.github/workflows/check-3rd-party.sh
+++ b/.github/workflows/check-3rd-party.sh
@@ -3,15 +3,14 @@
 FILE=$1
 status=0
 
-while read -r line; do
-  src=`echo $line | sed -E "s/(.*)uses://g" | xargs`
-  author=`echo $src | awk -F/ '{print $1}'`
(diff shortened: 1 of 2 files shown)


.github/workflows/check-3rd-party.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+FILE=$1
+status=0
+
+while read -r line; do
+  src=`echo $line | sed -E "s/(.*)uses://g" | xargs`
+  author=`echo $src | awk -F/ '{print $1}'`
```
Stebalien, line 8: I'd check that this is actually github. That is, check URL prefixes (e.g. `github.com/actions`) instead of users.

.github/workflows/check-3rd-party.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+FILE=$1
+status=0
+
+while read -r line; do
+  src=`echo $line | sed -E "s/(.*)uses://g" | xargs`
```
Stebalien, line 7: Why the xargs?

Also, this is going to be pretty brittle. Can we guarantee, e.g., the `yq` command (or something that can read yaml)?

.github/workflows/check-3rd-party.sh
marten-seemann: GitHub Actions are not included by URL, but by username/repo.
I don’t think anyone can publish an action using the “actions” username, but if you really want to make it stricter, we could create a whitelist. I don’t think that this is worth the effort though.

Stebalien: Ah, I see.

.github/workflows/check-3rd-party.sh
mvdan: Another option would be a language that can do the yaml parsing and string manipulation all at once, like a small python or go program. I think this is ultimately up to @marten-seemann's choice, depending on what software is readily available and does not require extra downloading/installing.

marten-seemann: I was trying to avoid baking in the assumption that all workflows will be valid yaml. If we decide to use a templating system later, we might end up with something that yaml parser won't like.

Stebalien: Ok, fair enough. Maybe `sed -ne 's/[[:space:]]*uses:[[:space:]]*//p`? (can also skip the grep this way)

marten-seemann: Small modification: `sed -ne 's/[[:space:]-]*uses:[[:space:]]*//p'` also removes the (optional) `-`.

marten-seemann:
I had to add another `| sed -e 's/\s*#.*$//'` after the first `sed` to remove the version comment. This could probably done in a single `sed` comment, but I don't have the patience to figure out how.

.github/workflows/check-3rd-party.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+FILE=$1
+status=0
+
+for line in `sed -ne 's/[[:space:]-]*uses:[[:space:]]*//p' $1 | sed -e 's/\s*#.*$//'`; do
+  author=`echo $line | awk -F/ '{print $1}'`
+  if [ $author == "actions" ]; then continue; fi
```
mvdan, line 8: since this is bash, might as well use [[ instead of [, which has fewer footguns :)

.github/workflows/check-3rd-party.sh
```diff
@@ -0,0 +1,16 @@
...
+  version=`echo $line | awk -F@ '{print $2}' | awk '{print $1}'`
+  if [ ${#version} != 40 ]; then
```
mvdan, line 10: perhaps consider a regexp with [[, so that you don't accept a tag that just so happens to be 40 characters long: https://www.networkworld.com/article/2693361/unix-tip-using-bash-s-regular-expressions.html

You can never be sure, but at least you can require exactly 40 hexadecimal characters

mvdan (approved):

.github/workflows/check-3rd-party.sh
marten-seemann: Makes sense.

mvdan (approved):

PR #24, opened 2021-03-13
check go mod tidy

marten-seemann:


Stebalien (approved):

mvdan (approved):
Another option would be assuming git and using `git diff`. Usually that could be a problem if the modified file affects the rest of CI, but tidying can only fix things up, it shouldn't introduce any problems or side effects.

Longer term, I think https://github.com/golang/go/issues/27005 is the best solution, but it could be a year before that's fleshed out and released. You could add a TODO comment pointing to that issue as a reminder, though.

PR #25, opened 2021-03-18
make it possible to use this workflow with > 64 repos

marten-seemann:
## Problem

GitHub Actions limits the number of jobs in a matrix to 256. We create one job per repository and per file that we're copying. Therefore at the moment we're limited to 256/4 = 64 repositories.

## Solution

This PR uses the [repository dispatch](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#repository_dispatch) function. In a nutshell, this feature allows us to emit a custom event named `copy-workflow`. We can then define another workflow that's triggered by this event.

### Details

The `dispatch.yml` workflow parses the `config.json` file (which contains a list of all participating repos). It batches the repositories into batches of 50: https://github.com/ipld/.github/blob/adb77431ecc3416bd1635e1f5bbcc640e41405f0/.github/workflows/dispatch.yml#L24-L25

For example, if we had 500 participating repositories, we'd end up with 10 batches of 50 repos. For every one of those batches, we then trigger a `copy-workflow` event. We pass both the list of repositories in this batch as well as the list of files to that workflow:
https://github.com/ipld/.github/blob/adb77431ecc3416bd1635e1f5bbcc640e41405f0/.github/workflows/dispatch.yml#L38-L42.

Each `copy-workflow` then creates a matrix running one job per target repository and file.

### Limitations

The dispatch workflow also uses a matrix, which limits us to 256 batches. Theoretically, the upper limit of repositories supported is therefore 256*256/(num files) = 16,384 (for 4 files). In practice, this limit will be a bit lower once we add a options to `config.json` to configure which files will be synced for a particular repository. Still, this should be plenty, considering that currently "only" have O(1000) repos.

### DIFF
--- a/.github/workflows/dispatch.yml
+++ b/.github/workflows/dispatch.yml
@@ -1,0 +1,42 @@
+# Trigger the execution of copy-workflow.yml, in batches of 5 repositories.
+# This workflow is needed since GitHub Actions limits the matrix size to 256 jobs.
+# We use one job per repository per file.
+
+on:
+  pull_request:
+    branches: [ master ]
+    types: [ closed ] # IMPORTANT: check the merge status for every job: github.event.pull_request.merged == true
+
+env:
+  # We could use a higher number here. We use a small number just to make sure to create multiple batches.
+  MAX_REPOS_PER_WORKFLOW: 5
+  WORKFLOWS: '[ "autorebase", "automerge", "go-test", "go-check" ]' # a JSON array of the files to distribute
+
+jobs:
+  matrix:
+    if: github.event.pull_request.merged == true
+    name: Trigger copy workflows
+    runs-on: ubuntu-latest
+    outputs:
+      targets: ${{ steps.set-matrix.outputs.targets }}
+    steps:
+      - uses: actions/checkout@v2
+      - id: set-matrix
+        run: |
+          TARGETS=$(jq '. | _nwise(${{ env.MAX_REPOS_PER_WORKFLOW }})' .github/workflows/config.json | jq -sc '. | to_entries')
+          echo "::set-output name=targets::$TARGETS"
+  dispatch:
+    if: github.event.pull_request.merged == true
+    needs: [ matrix ]
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        cfg: ${{ fromJson(needs.matrix.outputs.targets) }}
+    name: Start copy workflow (batch ${{ matrix.cfg.key }})
+    steps:
+      - uses: peter-evans/repository-dispatch@ce5485de42c9b2622d2ed064be479e8ed65e76f4 # v1.1.3
+        with:
+          token: ${{ secrets.WEB3BOT_GITHUB_TOKEN }}
+          event-type: copy-workflow
+          client-payload: '{ "github": ${{ toJson(github) }}, "workflows": ${{ env.WORKFLOWS }}, "targets": ${{ toJson(matrix.cfg.value) }} }'
--- a/.github/workflows/copy-workflow.yml
+++ b/.github/workflows/copy-workflow.yml
@@ -1,29 +1,17 @@
+# This workflow is triggered by the dispatch workflow.
+
 on:
-  pull_request:
-    branches: [ master ]
-    types: [ closed ] # IMPORTANT: check the merge status for every job: github.event.pull_request.merged == true
+  repository_dispatch:
+    types: [ copy-workflow ]
 
 jobs:
-  matrix:
-    if: github.event.pull_request.merged == true
-    runs-on: ubuntu-latest
-    outputs:
-      targets: ${{ steps.set-matrix.outputs.targets }}
-    steps:
-      - uses: actions/checkout@v2
-      - id: set-matrix
-        run: |
-          TARGETS=$(jq -c . .github/workflows/config.json)
-          echo "::set-output name=targets::$TARGETS"
   copy:
-    if: github.event.pull_request.merged == true
-    needs: [ matrix ]
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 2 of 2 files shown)


Stebalien:
Thinking about this a bit, can we just trigger an event against all of our repos and ask them to pull from this repo? That should let this scale perfectly.

.github/workflows/dispatch.yml
```diff
@@ -0,0 +1,42 @@
...
+    name: Trigger copy workflows
+    runs-on: ubuntu-latest
+    outputs:
+      targets: ${{ steps.set-matrix.outputs.targets }}
+    steps:
+      - uses: actions/checkout@v2
+      - id: set-matrix
+        run: |
+          TARGETS=$(jq '. | _nwise(${{ env.MAX_REPOS_PER_WORKFLOW }})' .github/workflows/config.json | jq -sc '. | to_entries')
+          echo $TARGETS
```
Stebalien, line 26: did we mean to keep this line?

Stebalien (commented):
LGTM but I don't really know enough about github actions to be sure so I'd appreciate another review by someone else.

.github/workflows/dispatch.yml
```diff
@@ -0,0 +1,42 @@
...
+# This workflow is needed since GitHub Actions limits the matrix size to 256 jobs.
+# We use one job per repository per file.
+
+on:
+  pull_request:
+    branches: [ master ]
+    types: [ closed ] # IMPORTANT: check the merge status for every job: github.event.pull_request.merged == true
+
+env:
+  MAX_REPOS_PER_WORKFLOW: 50
```
marten-seemann, line 11: I think we should reduce this number to something smaller than the number of repos we currently have in our `config.json`, just to make sure that we end up with multiple batches to exercise the mechanism defined here?

.github/workflows/dispatch.yml
marten-seemann: Nope. Will remove.

marten-seemann:
> Thinking about this a bit, can we just trigger an event against all of our repos and ask them to pull from this repo?

I guess one could make that work, but I'm not sure if it simplifies things. The reason is that we'd have 2 different copy workflows: Every repository would need a `fetch-workflows` workflow that fetches `go-test`, `go-check` etc. from here. This workflow could also fetch updates to itself.
However, this doesn't work for the initial distribution. We'd need a separate workflow that just distributes `fetch-workflows` to repositories newly added to `config.json`.

Stebalien:
Yeah, you're right. And we can always revisit it later if it's a problem.

marten-seemann:
@mvdan Can I ask you to review this PR?

marten-seemann:
Note to self: If merged after #29, update the workflow to use the `WEB3BOT_GITHUB_TOKEN`.

Stebalien (approved):
@mikeal gave it an LGTM (but he didn't feel fully qualified to review).

I'd just merge this as, worst-case, we'll just have to fix it later.

.github/workflows/dispatch.yml
```diff
@@ -0,0 +1,41 @@
+# Trigger the execution of copy-workflow.yml, in batches of 50 repositories.
+# This workflow is needed since GitHub Actions limits the matrix size to 256 jobs.
+# We use one job per repository per file.
+
+on:
+  pull_request:
+    branches: [ master ]
+    types: [ closed ] # IMPORTANT: check the merge status for every job: github.event.pull_request.merged == true
```
mvdan, line 8: I'm surprised you can't just use "merged" mere :)

.github/workflows/dispatch.yml
```diff
@@ -0,0 +1,41 @@
...
+    if: github.event.pull_request.merged == true
+    needs: [ matrix ]
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        cfg: ${{ fromJson(needs.matrix.outputs.targets) }}
+    name: Start copy workflow (batch ${{ matrix.cfg.key }})
+    steps:
+      - uses: peter-evans/repository-dispatch@ce5485de42c9b2622d2ed064be479e8ed65e76f4 # v1.1.3
```
mvdan, line 37: is there a github forum thread or issue about doing this without a third party action? if so, it would be good to add it as a TODO comment, so we use more standard features longer-term

.github/workflows/dispatch.yml
```diff
@@ -0,0 +1,41 @@
...
+        with:
+          token: ${{ secrets.WEB3BOT_GITHUB_TOKEN }}
+          event-type: copy-workflow
+          client-payload: '{ "github": ${{ toJson(github) }}, "workflows": ${{ env.WORKFLOWS }}, "targets": ${{ toJson(matrix.cfg.value) }} }'
```
mvdan, line 41: It worries me that we're passing the entire `github` object. Even if we make sure it doesn't contain sensitive information right now, perhaps it will in the future.

You seem to only need the event data, so how about `"event": ${{ toJson(github.event) }}`?

mvdan (approved):
LGTM; I agree with @marten-seemann's investigation on how best to do this right now.

I imagine GitHub will eventually improve the built-in ways for one job to trigger others, even across repositories.

mvdan:
Also, apologies for the slowness here, as I was out unexpectedly last week.

.github/workflows/dispatch.yml
marten-seemann: Reduced to 5.

.github/workflows/dispatch.yml
marten-seemann: I haven't seen any. As far as I can tell, GitHub seems to be happy if things can be solved using 3rd party actions.

.github/workflows/dispatch.yml
marten-seemann: Sorry, I forgot to address your comment. Will open an issue.

Issue #28, opened 2021-03-24
Auto-deploy dependabot config


Stebalien:
Dependabot is now configured through `.github/dependabot.yml`. We should deploy this in at least _some_ of our repos.

marten-seemann:
Can you paste a sample config file here? Do we need this config file in all participating (Go) repos, or do we need a config option for that in `config.json`?

Stebalien:
`.github/dependabot.yml`

```yaml
version: 2
updates:
- package-ecosystem: gomod
  directory: "/"
  schedule:
    interval: daily
    time: "11:00"
  open-pull-requests-limit: 10
  labels:
    - "topic/dependencies"

```

And yeah... we need a config file in all participating go repos.

marten-seemann:
> And yeah... we need a config file in all participating go repos.

Good, that simplifies things a bit for now.

What doesn't simplify things is that the path is `.github/dependabot.yml`. Currently, the copy workflow assumes that all files go to `.github/workflows/`.
I think we should change the directory structure here a bit, to make it work in the general case. We could rewrite the workflow such that it copies all files from `templates/A/B/C.xyz` in this repo to `A/B/C.xyz` in the target repo.

Stebalien:
Yeah, that sounds like a good idea.

galargh:
We're going to revisit this as part of https://github.com/pl-strflt/ipdx/issues/91. Now that we have more tools at our disposal, more configurable Unified CI - https://github.com/pl-strflt/uci - and GitHub Management - https://github.com/protocol/github-mgmt-router - it should be easier to carry out.

PR #30, opened 2021-03-27
add issue label config

Stebalien:
for use with https://github.com/marketplace/actions/label-sync
related to https://github.com/protocol/web3-dev-team/pull/90

### DIFF
--- a/scripts/sync-labels/.gitignore
+++ b/scripts/sync-labels/.gitignore
@@ -1,0 +1,1 @@
+node_modules/
--- a/scripts/sync-labels/README.md
+++ b/scripts/sync-labels/README.md
@@ -1,0 +1,25 @@
+# Issue Label Syncer
+
+Manually sync issue labels.
+
+## Install
+
+Before usage, you'll need to install the dependencies.
+
+```go
+$ npm install
+```
+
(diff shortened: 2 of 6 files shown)


marten-seemann (commented):
This PR seems incomplete, as it only adds the config file, but no instruction to actually copy the file (which, of course, is blocked by the fact that we currently can only copy files in `.github/workflows`, see https://github.com/protocol/.github/issues/28#issuecomment-807870697).
Can you convert it to a draft PR?

Stebalien:
Yep, still working on it.

galargh:
Hey! If we're still interested in labels synchronisation, I think we should do this through GitHub Management repos. It doesn't have native label support yet but it's planned - https://github.com/protocol/github-mgmt-template/issues/81. 

In the meantime, we can create custom scripts to deal with the lack of native support. Here is an example of how we sync labels in `js-libp2p` and `js-libp2p-peer-store` repos https://github.com/libp2p/github-mgmt/blob/master/scripts/src/actions/sync-labels.ts.

I'm going to close this since there was no activity for a long time but please do let me know if you wanted set up label syncing in any repos, I'd be happy to assist.

PR #32, opened 2021-03-27
add a few libp2p and multiformats repositories

marten-seemann:


Stebalien (approved):

Stebalien:
FYI, I now have a script in #30 that you can use to just scrape the entire repo list when we finally pull that trigger.

marten-seemann:
> FYI, I now have a script in #30 that you can use to just scrape the entire repo list when we finally pull that trigger.

Nice! We could write a cron workflow that automatically adds newly created repos (at some point in the future).

PR #35, opened 2021-03-27
don't restrict the copy workflow to .github/workflows/

marten-seemann:


templates/header.yml
```diff
@@ -1,3 +1,3 @@
-# Workflow managed by web3-bot. DO NOT EDIT.
+# File managed by web3-bot. DO NOT EDIT.
```
marten-seemann, line 1: Sorry for the churn here. This will trigger an update of all workflow files :(

.github/workflows/copy-workflow.yml
```diff
@@ -14,10 +14,11 @@ jobs:
         workflow: ${{ github.event.client_payload.workflows }}
     env:
       WORKFLOW_DIR: "workflow-repo"
-      FILE: "${{ matrix.workflow }}.yml"
+      TEMPLATE_DIR: "templates"
+      FILE: "${{ matrix.workflow }}"
```
mvdan, line 18: should all those other references to "workflow" or "WORKFLOWS" also be replaced with a generic "file" or "FILES"?

mvdan (approved):

.github/workflows/copy-workflow.yml
marten-seemann: Yes, that makes sense.

PR #39, opened 2021-03-31
run gofmt in go-check

### LINKED ISSUE #26: run gofmt
`gofmt -e -d ./...` outputs the diff, which is nice, but it won't allow GitHub to annotate the PR.
It also always exits with status 0, so we'd have to look at the length of the output to actually fail the build on `gofmt` problems.

@mvdan, since you obviously spent a lot of time thinking about `gofmt`, what's the best way to run it in CI?

marten-seemann:
Fixes #26.

Sample output in https://github.com/marten-seemann-test/target/pull/53/files.

### DIFF
--- a/templates/.github/workflows/go-check.yml
+++ b/templates/.github/workflows/go-check.yml
@@ -18,6 +18,14 @@ jobs:
           go mod tidy
           diff go.mod go.mod.orig
           diff go.sum go.sum.orig
+      - name: gofmt
+        if: ${{ success() || failure() }} # run this step even if the previous one failed
+        run: |
+          out=$(gofmt -s -l .)
+          if [[ -n "$out" ]]; then
+            echo $out | awk '{print "::error file=" $0 ",line=0,col=0::File is not gofmt-ed."}'
+            exit 1
+          fi
       - name: go vet
         if: ${{ success() || failure() }} # run this step even if the previous one failed
         run: go vet ./...


templates/.github/workflows/go-check.yml
```diff
@@ -18,6 +18,14 @@ jobs:
           go mod tidy
           diff go.mod go.mod.orig
           diff go.sum go.sum.orig
+      - name: go fmt
+        if: ${{ success() || failure() }} # run this step even if the previous one failed
+        run: |
+          out=$(gofmt -l .)
```
mvdan, line 24: consider adding -s; if you want to leave it out for now, add it as a TODO

templates/.github/workflows/go-check.yml
```diff
@@ -18,6 +18,14 @@ jobs:
           go mod tidy
           diff go.mod go.mod.orig
           diff go.sum go.sum.orig
+      - name: go fmt
```
mvdan, line 21: nit: gofmt :) otherwise it might be surprising to the end user that Go files outside of Go packages are being formatted too.

mvdan (approved):

templates/.github/workflows/go-check.yml
marten-seemann: Makes sense, as long as we use it in the automation as well (which I'll make sure we do).

Stebalien (approved):

PR #42, opened 2021-04-01
run go mod tidy (and set Go 1.15) and gofmt -s in copy workflow

marten-seemann:


Stebalien (approved):

Stebalien:
Nice! We could also consider running `go fix` later, but we can do that if we see something in the go release notes that might require it.

.github/workflows/copy-workflow.yml
```diff
@@ -23,14 +23,33 @@ jobs:
...
+      run: |
+        git config user.name web3-bot
+        git config user.email "web3-bot@users.noreply.github.com"
+    - name: run go mod tidy
+      run: |
+        go mod edit -go 1.15
+        go mod tidy
+        if ! git diff --quiet; then
+          git add .
+          git commit -m "run go mod tidy"
```
mvdan, line 36: it might be confusing to a user that the go version was changed, which is not something that tidy would do by itself.

how about "set go version to 1.15 and go mod tidy"?

.github/workflows/copy-workflow.yml
```diff
@@ -23,14 +23,33 @@ jobs:
...
+          echo "NEEDS_UPDATE=1" >> $GITHUB_ENV
+        fi
+    - name: gofmt
+      run: |
+        gofmt -s -w .
+        if ! git diff --quiet; then
+          git add .
+          git commit -m "run gofmt -s"
+          echo "NEEDS_UPDATE=1" >> $GITHUB_ENV
```
mvdan, line 45: I kinda wonder if you could later just ask git "are there any commits ahead of the upstream branch?" instead of using NEEDS_UPDATE. Unrelated to this change though.

mvdan (approved):

.github/workflows/copy-workflow.yml
marten-seemann: As we're committing to master here, we could just check if HEAD != origin/master. Just need to figure out an elegant way to pull that out of git.

marten-seemann: Will do that in a separate PR.

.github/workflows/copy-workflow.yml
marten-seemann: Done.

.github/workflows/copy-workflow.yml
mvdan: Sounds good :) all LGTM

PR #43, opened 2021-04-01
add the script used check the Go repo health

marten-seemann:
Depends on #145, as it used the `get-repos` script.

This is the script that was used to generate the [Go Repo Health](https://docs.google.com/spreadsheets/d/1Vwk-OUKXXNQkp4BkOoMndLylnJRW7a2c_mmvXu6jIj8/edit#gid=0) spreadsheet, now with an added column indicating if web3-bot already created a PR or not.

### DIFF
--- a/scripts/health.sh
+++ b/scripts/health.sh
@@ -1,0 +1,63 @@
+#!/bin/bash
+
+to_status() {
+  if [[ $1 == 0 ]]; then
+    printf "✅"
+    return
+  fi
+  printf "❌"
+}
+
+check_go_mod_tidy() {
+  if [[ ! -f go.mod || ! -f go.sum ]]; then
+    echo 1
+    return
+  fi
+  cp go.mod go.mod.orig
+  cp go.sum go.sum.orig
+  go mod tidy &> /dev/null
+  diff go.mod go.mod.orig > /dev/null
+  if [[ $? != 0 ]]; then
+    echo 1
+    return
+  fi
+  diff go.sum go.sum.orig > /dev/null
+  echo $?
+}
+
+check_go_fmt() {
+  out=$(gofmt -l . | xargs)
+  if [[ -z "$out" ]]; then
+    echo 0
+    return
+  fi
+  echo 1
+}
+
+for repo in $(node get-repos/index.js -l Go | jq -r '.[]'); do
+  tmp=$(mktemp -d)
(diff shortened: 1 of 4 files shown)


scripts/health.sh
```diff
@@ -0,0 +1,63 @@
...
+  cp go.mod go.mod.orig
+  cp go.sum go.sum.orig
+  go mod tidy &> /dev/null
+  diff go.mod go.mod.orig > /dev/null
+  if [[ $? != 0 ]]; then
+    echo 1
+    return
+  fi
+  diff go.sum go.sum.orig > /dev/null
+  echo $?
```
Stebalien, line 25: why echo? We could just do what we do for the other commands (capture the return status)

scripts/health.sh
```diff
@@ -0,0 +1,63 @@
...
+    echo 0
+    return
+  fi
+  echo 1
+}
+
+for repo in $(node get-repos/index.js -l Go | jq -r '.[]'); do
+  tmp=$(mktemp -d)
+  pushd $tmp > /dev/null
+  git clone -q https://github.com/$repo .
```
Stebalien, line 40: so... it _would_ be nice if there were a way to tell this to clone from local repos... E.g., from a gopath.

Maybe just make the `https://github.com` part configurable?

scripts/health.sh
```diff
@@ -0,0 +1,63 @@
...
+      go vet ./... &> /dev/null
+      govet=$?
+      staticcheck ./... &> /dev/null
+      staticcheck=$?
+      echo "$repo $numcommits false $(to_status $gomodtidy) $(to_status $gofmt) $(to_status $gotest) $(to_status $gotestrace) $(to_status $govet) $(to_status $staticcheck)"
+    else
+      echo "$repo $numcommits true"
+    fi
+  fi
+  popd > /dev/null
```
Stebalien, line 60: I'd honestly just run this all in a subshell unless we _need_ it to be in the main shell. I.e., replace the push/pop with

```bash
(
...
)
```

Stebalien (approved):

galargh:
We haven't been using this script in a while, and we do try to cover new Go repos as soon as they're created, but it might be interesting to run it and see what's the situation with the repos that are yet not covered. I'm going to merge it as-is as it's not on a critical path.

PR #44, opened 2021-04-01
remove Travis and CircleCI configs on first deployment of test workflow

marten-seemann:
As suggested by @aschmahmann, when we first deploy the `go-test` workflow to repositories, we should disable Travis and CircleCI. The copy workflow adds commits deleting `.travis.yml` and `.circleci/`, IF `.github/workflows/go-test.yml` doesn't exist yet (and thus will be added in the same PR).


mvdan (approved):
I think three steps is a bit overkill though. This could be a single step with three conditionals.

marten-seemann:
> I think three steps is a bit overkill though. This could be a single step with three conditionals.

True, but steps are cheap. And this way we get a nice visual indication if a step was executed:
image

PR #52, opened 2021-04-03
run dispatch workflow when pushing to master

### LINKED ISSUE #49: Can't deploy PRs merged from forks
https://github.com/protocol/.github/actions/runs/712309930

marten-seemann:
Fixes #49.

Our current logic is slightly wrong: We build the (last) commit of a PR, not the merge commit. We did this because it allows us easy access to the PR title, which GitHub doesn't give us for a push to master.
Furthermore, since we were building a PR, GitHub would only give us access to the repository's secrets if we weren't building from a fork.

This PR changes the logic: We now build every push on master. As master is branch-protected, this means we'll only build merge commits. We lose access to the PR title here, which is why I replaced the PR title with a generic "sync: update CI config files" (which @mvdan suggested anyway).

As a future improvement, we might be able to regain access to some of the information if we restrict the merge method to squashing, but I haven't tried this out yet.

### DIFF
--- a/.github/workflows/copy-workflow.yml
+++ b/.github/workflows/copy-workflow.yml
@@ -102,9 +102,10 @@ jobs:
       if: ${{ env.NEEDS_UPDATE == 1 }}
       uses: peter-evans/create-pull-request@052fc72b4198ba9fbc81b818c6e1859f747d49a8 #v3.8.2
       with:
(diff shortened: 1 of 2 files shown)


mvdan (approved):
I think this is much better overall.

Stebalien (approved):

PR #54, opened 2021-04-04
always upload coverage report to Codecov

marten-seemann:
We recently turned down Codecov's verbosity by using the config suggested by the [Go Best Practice document](https://github.com/ipld/ipld/blob/master/best-practices/go-ci.md#using-code-coverage-tools). We can now afford to upload a coverage report for every test run.


mvdan (approved):
I guess the action is a no-op if no codecov config exists?

PR #59, opened 2021-04-07
make it possible to use different GitHub users per GitHub org

marten-seemann:
IPFS and Filecoin already use separate GitHub users, which Andy created earlier today.
If everything works out, we'll also create separate GitHub users for all the other orgs. With this PR in place, all we have to do is update the user configuration in `github-users.json`.

### DIFF
--- a/.github/workflows/copy-workflow.yml
+++ b/.github/workflows/copy-workflow.yml
@@ -12,6 +12,7 @@ jobs:
       matrix:
         cfg: ${{ github.event.client_payload.targets }}
     env:
+      TARGET_REPO_DIR: "target-repo"
       TEMPLATE_REPO_DIR: "template-repo"
       TEMPLATE_DIR: "templates"
       NEEDS_UPDATE: 0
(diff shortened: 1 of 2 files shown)


.github/workflows/copy-workflow.yml
```diff
@@ -12,43 +12,62 @@ jobs:
       matrix:
         cfg: ${{ github.event.client_payload.targets }}
     env:
+      TARGET_REPO_DIR: "target-repo"
       TEMPLATE_REPO_DIR: "template-repo"
       TEMPLATE_DIR: "templates"
       NEEDS_UPDATE: 0
       INITIAL_TEST_DEPLOYMENT: 0
+      GITHUB_USER: ""
+      GITHUB_EMAIL: ""
```
mvdan, line 21: I imagine this still works without these zero initial values? The same applies to NEEDS_UPDATE and others.

.github/workflows/github-users.json
```diff
@@ -0,0 +1,23 @@
...
+    "user": "web3-bot-filecoin"
+  },
+  "filecoin-shipyard": {
+    "user": "web3-bot"
+  },
+  "ipfs": {
+    "user": "web3-bot-ipfs"
+  },
+  "ipfs-cluster": {
+    "user": "web3-bot"
```
mvdan, line 12: naively, wouldn't this be web3-bot-ipfs?

.github/workflows/github-users.json
```diff
@@ -0,0 +1,23 @@
+{
+  "filecoin-project": {
+    "user": "web3-bot-filecoin"
+  },
+  "filecoin-shipyard": {
+    "user": "web3-bot"
```
mvdan, line 6: same, but for web3-bot filecoin?

.github/workflows/copy-workflow.yml
```diff
@@ -12,43 +12,62 @@ jobs:
       matrix:
         cfg: ${{ github.event.client_payload.targets }}
     env:
+      TARGET_REPO_DIR: "target-repo"
```
mvdan, line 15: not sure I understand why we need a sub-directory now. maybe document that need?

mvdan (approved):

.github/workflows/github-users.json
marten-seemann: That depends on how Andy configures it. At the moment there aren't any changes. Later on, it's debatable if this should be web3-bot-ipfs or web3-bot-ipfs-cluster. For consistency, I assume we'll probably just create a new user.

.github/workflows/copy-workflow.yml
marten-seemann: Previously, we used to clone the template repository into a subdirectory of the target directory. Which works fine, until you do a `git add .`. This wasn't a problem until now, because we just deferred cloning the template directory until we had all the commits in place.
With this change, this is not possible any more, since we need the `user-config.json` file from the template directory before we create any commits.

.github/workflows/copy-workflow.yml
marten-seemann: Unfortunately it doesn't, I tried it. GitHub Actions seems to keep environment variables tightly scoped: If you don't declare it for the job, it will only be available in the step where you set it.

PR #60, opened 2021-04-08
remove the autorebase workflow

marten-seemann:
For security reasons, we shouldn't give every repo access to a WEB3BOT_GITHUB_TOKEN that's valid for the entire org. Unfortunately, the `autorebase` workflow relies on a Personal Access Token: https://github.com/cirrus-actions/rebase#installation.
Using the regular GITHUB_TOKEN (which would be available to every workflow) only allows us to rebase the PR, but it wouldn't trigger any other actions (in particular, it doesn't re-run `go-test` and `go-check` on the rebased PR).


.github/workflows/copy-workflow.yml
```diff
@@ -33,6 +33,12 @@ jobs:
         if [[ ! -f .github/workflows/go-test.yml ]]; then
           echo "INITIAL_TEST_DEPLOYMENT=1" >> $GITHUB_ENV
         fi
+    - name: remove the autorebase workflow
```
mvdan, line 36: perhaps add a TODO to remove this eventually? perhaps in a month's time, if we think all files are gone in master by then.

mvdan (approved):

PR #65, opened 2021-04-21
check that config.json is sorted alphabetically

marten-seemann:
`config.json` will be a long file (hundreds of lines). Make sure that it's sorted.

### DIFF
--- a/.github/workflows/check-config.yml
+++ b/.github/workflows/check-config.yml
@@ -10,7 +10,7 @@ jobs:
       - name: check if config.json is sorted alphabetically
         run: |
           jq -r ".[].target" config.json > entries.txt
-          sort entries.txt > entries_sorted.txt
+          sort -u entries.txt > entries_sorted.txt
           if ! output=$(diff -y entries.txt entries_sorted.txt); then
             echo "Targets in config.json not sorted alphabetically:"
             echo "$output"
--- a/.github/workflows/check-config.yml
+++ b/.github/workflows/check-config.yml
@@ -1,0 +1,18 @@
+on: [ pull_request ]
+
+jobs:
+  check-config:
+    runs-on: ubuntu-latest
+    outputs:
+      targets: ${{ steps.set-matrix.outputs.targets }}
+    steps:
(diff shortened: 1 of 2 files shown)


marten-seemann:
Sample output if unsorted:
image

.github/workflows/check-config.yml
```diff
@@ -0,0 +1,18 @@
...
+    runs-on: ubuntu-latest
+    outputs:
+      targets: ${{ steps.set-matrix.outputs.targets }}
+    steps:
+      - uses: actions/checkout@v2
+      - name: check if config.json is sorted alphabetically
+        run: |
+          jq -r ".[].target" config.json > entries.txt
+          sort entries.txt > entries_sorted.txt
+          if ! output=$(diff -y entries.txt entries_sorted.txt); then
```
mvdan, line 14: huh, how does this work? I guess because diff returns an exit code of 1 if there is a diff, and then the parent command sees that exit code too?

.github/workflows/check-config.yml
```diff
@@ -0,0 +1,18 @@
...
+  check-config:
+    runs-on: ubuntu-latest
+    outputs:
+      targets: ${{ steps.set-matrix.outputs.targets }}
+    steps:
+      - uses: actions/checkout@v2
+      - name: check if config.json is sorted alphabetically
+        run: |
+          jq -r ".[].target" config.json > entries.txt
+          sort entries.txt > entries_sorted.txt
```
mvdan, line 13: perhaps use `sort -u`, which will also make this check complain if there are duplicate entries.

mvdan (approved):

.github/workflows/check-config.yml
marten-seemann: `diff -y` always returns an output, but we only want to print it when there's actually a diff between the two files.

.github/workflows/check-config.yml
marten-seemann: Good point, I hadn't thought of duplicates.

PR #66, opened 2021-04-21
make it possible to run the check-config script as a pre-commit hook

marten-seemann:
Let's do it the fancy way ;)

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -1,0 +1,12 @@
+#!/bin/bash
+
+tmp=$(mktemp)
+git show :config.json > $tmp
+status=0
+if ! output=$(.github/workflows/check-config.sh $tmp); then
+  echo "$output"
+  status=1
+fi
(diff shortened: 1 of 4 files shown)


.githooks/pre-commit
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+tmp=$(mktemp)
+git show :config.json > $tmp
+status=0
```
marten-seemann, line 5: I wish there was a way to just say `defer rm $tmp`.

mvdan: you technically can via a trap, but it's not particularly pretty either: http://redsymbol.net/articles/bash-exit-traps/

mvdan (approved):
I wonder if this is a bit overkill, but it can't hurt :)

marten-seemann:
> I wonder if this is a bit overkill, but it can't hurt :)

It definitely is ;)

PR #67, opened 2021-04-21
add more libp2p and multiformats repos

marten-seemann:
For these repos, according to https://docs.google.com/spreadsheets/d/1Vwk-OUKXXNQkp4BkOoMndLylnJRW7a2c_mmvXu6jIj8, all tests and checks should pass.

Blocked by the permission issue that currently prevents the dispatch-workflow from executing.


mvdan (approved):
SGTM. go-multicodec is the only one I know here and it's of course fine to add.

PR #69, opened 2021-04-22
remove GitHub user config

marten-seemann:
We decided to use web3-bot for all orgs.


.github/workflows/copy-workflow.yml
```diff
@@ -17,8 +17,8 @@ jobs:
       TEMPLATE_DIR: "templates"
       NEEDS_UPDATE: 0
       INITIAL_TEST_DEPLOYMENT: 0
-      GITHUB_USER: ""
-      GITHUB_EMAIL: ""
+      GITHUB_USER: "web3-bot"
+      GITHUB_EMAIL: "web3-bot@users.noreply.github.com"
```
marten-seemann, line 21: We never change these envs, but I guess it doesn't hurt to use them as constants.

Stebalien (approved):

PR #74, opened 2021-04-22
don't require go.sum in go-check workflow

### LINKED ISSUE #73: go-check workflow fails if the repo doesn't have a go.sum
... and apparently it only needs one if there are entries in `go.sum`.

marten-seemann:
Fixes #73.

If `go.mod` doesn't contain any dependencies, it's not necessary to have a `go.sum` file.

Unfortunately, this change will trigger a PR in all repos that we've deployed to so far.

### DIFF
--- a/templates/.github/workflows/go-check.yml
+++ b/templates/.github/workflows/go-check.yml
@@ -13,11 +13,12 @@ jobs:
         run: go install honnef.co/go/tools/cmd/staticcheck@be534f007836a777104a15f2456cd1fffd3ddee8 # v2020.2.2
       - name: Check that go.mod is tidy
         run: |
-          cp go.mod go.mod.orig
-          cp go.sum go.sum.orig
           go mod tidy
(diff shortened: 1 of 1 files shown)


Stebalien (approved):

marten-seemann:
Waiting for @mvdan's approval before merging. He's had valuable input for bash and git in the past.

mvdan (approved):
I'm pretty sure `tidy` always adds a go.sum these days, even if it's empty. I assume this is about not failing in `cp go.sum go.sum.orig` if the file doesn't exist in the main branch yet.

marten-seemann:
>  I assume this is about not failing in cp go.sum go.sum.orig if the file doesn't exist in the main branch yet.

That's correct.

> I'm pretty sure tidy always adds a go.sum these days, even if it's empty.

It doesn't.

mvdan:
> It doesn't.

Huh, you're right, I'm not sure where I got that idea from.

PR #77, opened 2021-04-22
fix indent in copy workflow

marten-seemann:
Delaying merge until #74 is merged, in order to minimize the number of PRs we create.


mvdan (approved):
I absolutely love it when languages infer meaning from indentation :sob:

PR #84, opened 2021-04-30
actions: enhance actions

Stebalien:
1. Name them. This means the status checks get nice names instead of filenames.
2. Split go checks into "staticcheck" and "basic" so we can run them in parallel.

### DIFF
--- a/.github/workflows/check-3rd-party.yml
+++ b/.github/workflows/check-3rd-party.yml
@@ -1,5 +1,7 @@
 on: [ pull_request ]
 
+name: Check 3rd Party
+
 jobs:
   matrix:
     runs-on: ubuntu-latest
@@ -18,7 +20,7 @@ jobs:
       fail-fast: false
       matrix:
         file: ${{ fromJson(needs.matrix.outputs.targets) }}
-    name: check ${{ matrix.file }}
+    name: ${{ matrix.file }}
     steps:
       - uses: actions/checkout@v2
       - name: Run check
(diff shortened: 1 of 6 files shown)


marten-seemann (commented):
Is staticcheck that slow that we’ll gain something from parallelizing?
I’d expect that we’re limited by the number of GH Actions runners (10) in most cases.

Stebalien:
> Is staticcheck that slow that we’ll gain something from parallelizing?

Eh, probably not. I'll revert that.

Stebalien:
done

marten-seemann (approved):
LGTM. We might want to hold off merging this until we have more changes though, as this will trigger a PR to all repos.

Stebalien:
Yeah, about that... what if we did that on tag or something? Or had some kind of trigger?

Stebalien:
Eh, actually, we could also just PR into a "develop" branch, then merge that to master to deploy.

marten-seemann:
> Yeah, about that... what if we did that on tag or something? Or had some kind of trigger?

A tag doesn't work, since we're triggering on commits to master, and we don't have any information about the PR being merged at that point.

A trigger would be nice. We could use a custom event to run the dispatch workflow, but then we'd have to do a POST request via curl or so to trigger it.

> Eh, actually, we could also just PR into a "develop" branch, then merge that to master to deploy.

That might work. Changes to the workflow files would then go to the develop branch, and changes to `config.json` to master directly.

PR #85, opened 2021-05-02
only automerge if all commits were authored by web3-bot

marten-seemann:
That way, automerge won't trigger if:
* we added commits to the pull request that web3-bot opened, or
* we rebased the branch.

### DIFF
--- a/templates/.github/workflows/automerge.yml
+++ b/templates/.github/workflows/automerge.yml
@@ -4,9 +4,32 @@
 on: [ pull_request ]
 
 jobs:
-  automerge:
+  automerge-check:
     if: github.event.pull_request.user.login == 'web3-bot'
     runs-on: ubuntu-latest
+    outputs:
+      status: ${{ steps.should-automerge.outputs.status }}
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+    - name: Check if we should automerge
+      id: should-automerge
+      run: |
+        for commit in $(git rev-list --first-parent origin/${{ github.event.pull_request.base.ref }}..${{ github.event.pull_request.head.sha }}); do
+          committer=$(git show --format=$'%ce' -s $commit)
+          echo "Committer: $committer"
+          if [[ "$committer" != "web3-bot@users.noreply.github.com" ]]; then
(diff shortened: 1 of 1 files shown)


templates/.github/workflows/automerge.yml
```diff
@@ -4,9 +4,32 @@
...
-  automerge:
+  automerge-check:
     if: github.event.pull_request.user.login == 'web3-bot'
     runs-on: ubuntu-latest
+    outputs:
+      status: ${{ steps.should-automerge.outputs.status }}
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
```
mvdan, line 15: does this fetch all commits? I'd document that, because for a second I thought it meant zero commits.

mvdan (approved):

templates/.github/workflows/automerge.yml
marten-seemann: Yes, this is a non-shallow fetch. We need to do this, otherwise we don't know where `master` is.

marten-seemann: or rather, `github.event.pull_request.base.ref`

templates/.github/workflows/automerge.yml
```diff
@@ -4,9 +4,32 @@
...
+    - name: Check if we should automerge
+      id: should-automerge
+      run: |
+        for commit in $(git rev-list --first-parent origin/${{ github.event.pull_request.base.ref }}..${{ github.event.pull_request.head.sha }}); do
+          committer=$(git show --format=$'%ce' -s $commit)
+          echo "Committer: $committer"
+          if [[ "$committer" != "web3-bot@users.noreply.github.com" ]]; then
```
Stebalien, line 22: just so we're on the same page, this is a convenience feature (so we don't accidentally merge things before we're ready) not a security feature.

Stebalien (approved):

templates/.github/workflows/automerge.yml
marten-seemann: Yes. Security is dependent on the `if: github.event.pull_request.user.login == 'web3-bot'`.

While git commit author / committer can trivially be faked by an attacker, GitHub won't allow a user to open a PR for another user.

PR #88, opened 2021-05-05
chore: add more multiformats repos

marten-seemann:
Depends on https://github.com/multiformats/go-multiaddr-fmt/pull/15.


marten-seemann:
https://github.com/multiformats/go-multiaddr-fmt/pull/18 was merged, so this PR should be good to go.

PR #90, opened 2021-05-10
add more IPFS repos

marten-seemann:
These are the repos that @iand fixed last week.


Stebalien (approved):

config.json
```diff
@@ -23,11 +25,13 @@
   { "target": "ipfs/go-ipld-legacy" },
   { "target": "ipfs/go-merkledag" },
   { "target": "ipfs/go-metrics-prometheus" },
+  { "target": "ipfs/go-namesys" },
   { "target": "ipfs/go-path" },
   { "target": "ipfs/go-todocounter" },
   { "target": "ipfs/go-unixfsnode" },
   { "target": "ipfs/go-verifcid" },
   { "target": "ipfs/hang-fds" },
+  { "target": "ipfs/ipfs-update" },
```
Stebalien, line 34: Unfortunately, most of the tests for this project are "sharness" (run with `cd sharness && make test`). We'll need to add a special github action to run those.

config.json
```diff
@@ -3,11 +3,13 @@
   { "target": "filecoin-project/go-fil-commp-hashhash" },
   { "target": "filecoin-project/go-hamt-ipld" },
   { "target": "ipfs/go-block-format" },
+  { "target": "ipfs/go-cid" },
   { "target": "ipfs/go-detect-race" },
   { "target": "ipfs/go-ds-badger" },
   { "target": "ipfs/go-ds-bitcask" },
   { "target": "ipfs/go-ds-measure" },
   { "target": "ipfs/go-ds-pebble" },
+  { "target": "ipfs/go-ds-s3" },
```
Stebalien, line 12: These tests need access to a fake S3 service (we use a docker container) and a go-ipfs binary.

See https://github.com/ipfs/go-ds-s3/blob/master/.circleci/config.yml

I'm not sure how best to fix this. Maybe we add something to this config so we can disable the tests action for some repos? That way we can implement custom tests when necessary?

marten-seemann: This brings us back to templating, right? cc @mvdan

config.json
marten-seemann: https://github.com/ipfs/ipfs-update/pull/141

config.json
Stebalien: Note: I have a feeling we may need to switch to a yaml config to make that happen (at least without going insane). That is, I expect we'll want to be able to "patch-in" steps.

Stebalien: Yes, but I guess there are a few options:

1. We could just have some flags. In this case: "needs-s3: true".
2. Arbitrary yaml (in which case I'd switch this to config.yml so we can more easily configure it).

Option 1 is likely simpler and a better option at the moment.

Also, thinking about this, we _don't_ need the go-ipfs binary here. We only needed that for the plugin test, which should honestly be a separate action.

PR #95, opened 2021-05-17
add a cron workflow to find deleted repositories

marten-seemann:
Turns out we already have one deleted repo in our list (ipfs/go-ipfs-blockutil), and one archived one (ipfs/go-ipfs-cmdkit).
Sample run: https://github.com/protocol/.github/runs/2604478580?check_suite_focus=true

Future steps:
- Run this workflow on `pull_request` as well, but only only on newly added repos (if any).
- Create a PR that fixes `config.json` (this is probably overkill, we don't archive repos that often).

### DIFF
--- a/.github/workflows/stale-repos.yml
+++ b/.github/workflows/stale-repos.yml
@@ -1,0 +1,23 @@
+on:
+  schedule:
+  - cron: "0 */6 * * *" # every 6 hours
+
+jobs:
+  matrix:
+    name: Find stale repositories
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: find deleted repositories
+        run: |
+          status=0
+          for repo in $(jq -r '.[].target' config.json); do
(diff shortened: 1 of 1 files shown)


.github/workflows/stale-repos.yml
```diff
@@ -0,0 +1,23 @@
...
+    name: Find stale repositories
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: find deleted repositories
+        run: |
+          status=0
+          for repo in $(jq -r '.[].target' config.json); do
+            tmp=$(mktemp -d)
+            if ! git clone -q --depth 1 "https://github.com/$repo" "$tmp" 2> /dev/null; then
```
marten-seemann, line 16: Maybe we should just curl the URL and check for a 404?

marten-seemann: Resolved by calling the GitHub API endpoint.

mvdan (approved):

PR #102, opened 2021-05-30
add support for repositories with multiple Go modules

marten-seemann:
Closes #89.

Note: Codecov coverage reports are only uploaded for the main repo. Fixing this will be more difficult.

### DIFF
--- a/templates/.github/workflows/go-check.yml
+++ b/templates/.github/workflows/go-check.yml
@@ -12,7 +12,7 @@ jobs:
       - name: Install staticcheck
         run: go install honnef.co/go/tools/cmd/staticcheck@be534f007836a777104a15f2456cd1fffd3ddee8 # v2020.2.2
       - name: Check that go.mod is tidy
-        uses: protocol/multiple-go-modules@master
+        uses: protocol/multiple-go-modules@v1.0
         with:
           run: |
             go mod tidy
@@ -31,12 +31,12 @@ jobs:
           fi
       - name: go vet
         if: ${{ success() || failure() }} # run this step even if the previous one failed
-        uses: protocol/multiple-go-modules@master
+        uses: protocol/multiple-go-modules@v1.0
         with:
           run: go vet ./...
       - name: staticcheck
         if: ${{ success() || failure() }} # run this step even if the previous one failed
-        uses: protocol/multiple-go-modules@master
+        uses: protocol/multiple-go-modules@v1.0
         with:
           run: |
             set -o pipefail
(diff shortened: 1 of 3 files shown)


templates/.github/workflows/go-check.yml
```diff
@@ -12,27 +12,35 @@ jobs:
...
       - name: gofmt
         if: ${{ success() || failure() }} # run this step even if the previous one failed
-        run: |
-          out=$(gofmt -s -l .)
-          if [[ -n "$out" ]]; then
-            echo $out | awk '{print "::error file=" $0 ",line=0,col=0::File is not gofmt-ed."}'
-            exit 1
-          fi
+        uses: protocol/multiple-go-modules@master
+        with:
```
mvdan, line 27: not sure I agree with `gofmt -s -l .` running for each module. it walks directories, not modules. it's enough to just run it at the root of the repository. running it per-module will duplicate work if the modules are nested, and it might miss Go files outside of any module, which could happen with e.g. standalone programs with no external deps.

essentially, we want to use the multiple-go-modules action for any step that uses package patterns like `./...`, or for commands that directly interact with go.mod/go.sum like `go mod`, but nothing else.

mvdan (approved):

templates/.github/workflows/go-check.yml
marten-seemann: I had assumed that the Go tools would be consistent regarding their decision to work on a per-module, and not on a per-directory basis...

I'll update the PR.

templates/.github/workflows/go-check.yml
```diff
@@ -12,13 +12,15 @@ jobs:
...
         run: go install honnef.co/go/tools/cmd/staticcheck@be534f007836a777104a15f2456cd1fffd3ddee8 # v2020.2.2
       - name: Check that go.mod is tidy
-        run: |
-          go mod tidy
-          if [[ -n $(git ls-files --other --exclude-standard --directory -- go.sum) ]]; then
-            echo "go.sum was added by go mod tidy"
-            exit 1
-          fi
-          git diff --exit-code -- go.sum go.mod
+        uses: protocol/multiple-go-modules@master
```
marten-seemann, line 15: Should we cut a release (v1.0?) of multipe-go-modules?

Stebalien: Ys.

Stebalien (approved):

templates/.github/workflows/go-check.yml
mvdan: `go fmt`, with a space, does format packages. but most people should just use gofmt directly, as it also formats any Go files not matched bg `./...`.

PR #104, opened 2021-06-01
update staticcheck to 2021.1 (v0.2.0)

marten-seemann:
https://staticcheck.io/changes/2021.1


mvdan (approved):
This could mean changes needed in some repos, but I assume we want that.

marten-seemann:
Sample run (just to show that installing the package still works): https://github.com/libp2p/go-libp2p/runs/2720746306?check_suite_focus=true

> This could mean changes needed in some repos, but I assume we want that.

Yes.

PR #108, opened 2021-06-07
best-practices: port from the ipld/ipld repository with history

mvdan:


mvdan:
I used https://stackoverflow.com/a/11426261 for this, so that the original commit messages, authors, and dates, are kept.

I didn't change the path of the file as it still seemed OK.

best-practices/go-ci.md
```diff
@@ -0,0 +1,88 @@
...
+  status:
+    project:
+      default:
+        informational: true
+    patch:
+      default:
+        informational: true
+```
+
+#### How to use this document at Protocol Labs
```
marten-seemann, line 48: This section needs be heavily rewritten (or maybe removed?).

mvdan: Sure. Want to add an extra commit to this PR, for the sake of keeping the history intact?

PR #115, opened 2021-06-18
chore: add ipfs/tar-utils

marten-seemann:


aschmahmann (approved):
LGTM, thanks. Do I have to trigger the PR getting rebased on master if there's a change between when the PR goes out and it gets merged, or does that happen automatically?

marten-seemann:
> Do I have to trigger the PR getting rebased on master if there's a change between when the PR goes out and it gets merged, or does that happen automatically?

We'd have to do that manually, but if tests pass right away, we can just merge it immediately.

PR #139, opened 2021-07-22
upload coverage files for all modules to Codecov

### LINKED ISSUE #136: Upload coverage report of all go modules
If a go repo has multiple modules, it seems only the coverage for root module is uploaded. It would be fantastic to get coverage for all modules uploaded to codecov. 

Example:

`go-car` has [a module in root](https://github.com/ipld/go-car/blob/master/go.mod) and one in [v2](https://github.com/ipld/go-car/blob/master/v2/go.mod). But only the root module coverage shows up on [codecov](https://app.codecov.io/gh/ipld/go-car/).

marten-seemann:
Fixes #136.

### DIFF
--- a/templates/.github/workflows/go-test.yml
+++ b/templates/.github/workflows/go-test.yml
@@ -8,6 +8,8 @@ jobs:
       matrix:
         os: [ "ubuntu", "windows", "macos" ]
         go: [ "1.15.x", "1.16.x" ]
+    env:
+      COVERAGES: ""
     runs-on: ${{ matrix.os }}-latest
     name: ${{ matrix.os}} (go ${{ matrix.go }})
     steps:
@@ -37,6 +39,9 @@ jobs:
         uses: protocol/multiple-go-modules@v1.0
         with:
           run: go test -v -race ./...
+      - name: Collect coverage files
+        shell: bash
+        run: echo "COVERAGES=`find . -type f -name 'module-coverage.txt' | tr -s '\n' ',' | sed 's/.$//'`" >> $GITHUB_ENV
       - name: Upload coverage to Codecov
         uses: codecov/codecov-action@a1ed4b322b4b38cb846afb5a0ebfa17086917d27 # v1.5.0
         with:
--- a/templates/.github/workflows/go-test.yml
+++ b/templates/.github/workflows/go-test.yml
@@ -45,5 +45,5 @@ jobs:
       - name: Upload coverage to Codecov
         uses: codecov/codecov-action@a1ed4b322b4b38cb846afb5a0ebfa17086917d27 # v1.5.0
         with:
-          file: coverage.txt
+          files: ${{ env.COVERAGES }}
           env_vars: OS=${{ matrix.os }}, GO=${{ matrix.go }}
--- a/templates/.github/workflows/go-test.yml
+++ b/templates/.github/workflows/go-test.yml
@@ -26,7 +26,7 @@ jobs:
       - name: Run tests
         uses: protocol/multiple-go-modules@v1.0
         with:
-          run: go test -v -coverprofile coverage.txt ./...
+          run: go test -v -coverprofile module-coverage.txt ./...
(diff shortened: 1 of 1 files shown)


templates/.github/workflows/go-test.yml
```diff
@@ -37,6 +39,9 @@ jobs:
         uses: protocol/multiple-go-modules@v1.0
         with:
           run: go test -v -race ./...
+      - name: Collect coverage files
+        shell: bash
+        run: echo "COVERAGES=`find . -type f -name 'module-coverage.txt' | tr -s '\n' ',' | sed 's/.$//'`" >> $GITHUB_ENV
```
mvdan, line 44: I think you need to change -coverprofile to use module-coverage.txt now?

mvdan: FYI, best to use `$(cmd)` rather than `` `cmd` ``, as the latter syntax has been deprecated for a long time by the POSIX shell spec.

mvdan: One last nit: if the sed is to remove the trailing comma, I think `,$` would be more obvious.

marten-seemann: > I think you need to change -coverprofile to use module-coverage.txt now?

I had this in my test repo, but of course I forgot to apply it here... I'm glad I have reviewers who pay close attention :)

marten-seemann: Fixed and applied your suggestions.

mvdan (approved):
:tada:

masih (commented):
@marten-seemann Should the file name be updated [here](https://github.com/protocol/.github/blob/3bfa31bfe038792a25dcf7978d18814c2c5f42ed/templates/.github/workflows/go-test.yml#L48) too?

marten-seemann:
> @marten-seemann Should the file name be updated here too?

Good catch. Another copy-paste error 🤦‍♂️

templates/.github/workflows/go-test.yml
```diff
@@ -37,8 +39,11 @@ jobs:
...
         with:
           run: go test -v -race ./...
+      - name: Collect coverage files
+        shell: bash
+        run: echo "COVERAGES=$(find . -type f -name 'module-coverage.txt' | tr -s '\n' ',' | sed 's/,$//')" >> $GITHUB_ENV
       - name: Upload coverage to Codecov
         uses: codecov/codecov-action@a1ed4b322b4b38cb846afb5a0ebfa17086917d27 # v1.5.0
         with:
-          file: coverage.txt
+          files: ${{ env.COVERAGES }}
```
masih, line 48: We might want to quote paths considering usage here. I am wondering what would happen if a path to coverage file contains space.

masih (approved):

templates/.github/workflows/go-test.yml
masih: OK looks like the action [simply splits](https://github.com/codecov/codecov-action/blob/3375d62f3d177b734d1845da60213c1466b4f777/src/buildExec.ts#L96-L100) by `,`. In this case we might want to quote this entire value to assert the expanded value is passed as a string?

mvdan: this is yaml though, not shell - do you think spaces matter?

masih: Yep it's yaml. I am trying to think how it could break; directory names with `:` and other yaml special characters for example? The fundamental point here is, we mean to set the value of `files` key to string literal. So why not express it through syntax.

mvdan: Yeah I agree, a quoted string is probably a bit safer.

marten-seemann: Spaces should be fine, and I do hope that we don't use any weird special characters in directory names. But quoting doesn't hurt, so let's do it now before we run into any problems.

PR #166, opened 2021-08-10
chore: add ipld/codec-fixtures

marten-seemann:
We probably want to delay merging this PR until we merge the `next` branch (because there we also run `go mod tidy` in sub-packages).

### DIFF
--- a/config.json
+++ b/config.json
@@ -77,6 +77,7 @@
   { "target": "ipfs/iptb-plugins" },
   { "target": "ipfs/pinbot-irc" },
   { "target": "ipfs/tar-utils" },
+  { "target": "ipld/codec-fixtures" },
   { "target": "ipld/go-car" },
   { "target": "ipld/go-codec-dagpb" },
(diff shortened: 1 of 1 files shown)


mvdan (approved):

rvagg:
Does it matter at all that the source, and go.mod and go.sum are in a subdirectory? Are the standard scripts here going to find that?

marten-seemann:
It shouldn't matter, since we're using multiple-go-modules with all our Go-related commands, for example
https://github.com/protocol/.github/blob/8f4e735e0d7e99ec321334a7899da4a588e1e620/templates/.github/workflows/go-test.yml#L24-L27
multiple-go-modules searches all directories for go.mod files: https://github.com/protocol/multiple-go-modules/blob/a9324050387ecb413e92dac85a9def4bef505d46/action.yml#L19

The only exception is `gofmt`, but I don't think gofmt cares about the directory structure, it just processes all `*.go` files.

rvagg (approved):
:shrug: let's give it a go then eh?

PR #168, opened 2021-08-11
fix invalid quotes in copy workflow

marten-seemann:
Otherwise the copy workflow fails:
```
> Run for f in $(jq -r ".[]" <<< '"[ \".github/workflows/automerge.yml\", \".github/workflows/go-test.yml\", \".github/workflows/go-check.yml\" ]"'); do
jq: error (at <stdin>:1): Cannot iterate over string ("[ \".githu...)
```


marten-seemann:
It looks like it does, yes.

PR #172, opened 2021-08-13
make it possible to run repo-specific setup steps

marten-seemann:


README.md
```diff
@@ -5,6 +5,25 @@ By storing them in a central place (here), and distributing them in an automated
 1. Consistency: Every participating repository uses the same workflow, ensuring that our code adheres to the same coding standards and is properly tested.
 2. Maintainability: Workflows change over time. We need to be able to make changes without manually updating dozens of repositories.
 
+## Customization
+
+Most repositories won't need any customization, and the workflows defined here will just work fine.
+Some repositories may require some pre-setup steps to be run before tests (or code checks) can be run. Setup steps for `go-test` are defined in `.github/actions/go-test_setup/action.yml`, and setup steps for `go-check` are defined in `.github/actions/go-check_setup/action.yml`, in the following format:
+
+```yml
```
marten-seemann, line 13: Once this is used by any of our repos, we could link to that repo for a real-life example.

.github/workflows/check-3rd-party.sh
```diff
@@ -5,7 +5,7 @@ status=0
 
 for line in `sed -ne 's/[[:space:]-]*uses:[[:space:]]*//p' $1 | sed -e 's/\s*#.*$//'`; do
   author=`echo $line | awk -F/ '{print $1}'`
-  if [[ $author == "actions" || $author == "protocol" ]]; then continue; fi
+  if [[ $author == "." || $author == "actions" || $author == "protocol" ]]; then continue; fi
```
mvdan, line 8: document what the "dot" author is? at least I have no idea what that means

README.md
```diff
@@ -5,6 +5,25 @@ By storing them in a central place (here), and distributing them in an automated
 1. Consistency: Every participating repository uses the same workflow, ensuring that our code adheres to the same coding standards and is properly tested.
 2. Maintainability: Workflows change over time. We need to be able to make changes without manually updating dozens of repositories.
 
+## Customization
+
+Most repositories won't need any customization, and the workflows defined here will just work fine.
+Some repositories may require some pre-setup steps to be run before tests (or code checks) can be run. Setup steps for `go-test` are defined in `.github/actions/go-test_setup/action.yml`, and setup steps for `go-check` are defined in `.github/actions/go-check_setup/action.yml`, in the following format:
```
mvdan, line 11: I imagine a fairly common use case will be installing extra build dependencies, like for cgo, in which case the same step will be required for both. I guess we're fine with a bit of copy pasting or symlinking for that use case?

README.md
```diff
@@ -5,6 +5,25 @@ By storing them in a central place (here), and distributing them in an automated
...
+```yml
+runs:
+  using: "composite"
+  steps:
+    - name: Step 1 
+      shell: bash
+      run: echo "do some initial setup"
+    - name: Step 2
+      shell: bash
+      run: echo "do some more advanced setup"
```
mvdan, line 22: I think this should also demo a package install on Linux with `if: ...`, as I think that will be fairly common

templates/.github/workflows/go-check.yml
```diff
@@ -13,6 +13,9 @@ jobs:
         with:
           stable: 'false'
           go-version: "1.17.0-rc1"
+      - name: Run repo-specific setup
+        uses: ./.github/actions/go-check_setup
```
mvdan, line 17: nit: mixing dashes and underscores feels prone to mistakes. How about `go-check-setup`? It's not like we're being consistent with using underscores, either.

mvdan (approved):

README.md
marten-seemann: I think we are.

Another common case would be installing and running a background job (IPFS daemon), which we'll only need for `go-test`, which I'd like to avoid for `go-check`.

README.md
marten-seemann: I didn't expect it to work, but apparently we actually have access to the `matrix` variable in the local action, see https://github.com/protocol/.github-test-target/runs/3321274123.
I updated the README.

templates/.github/workflows/go-check.yml
marten-seemann: I thought it would be nice to have a different separator, in case at some point we introduce a `go-check_teardown`. No strong feelings either way though.

.github/workflows/check-3rd-party.sh
marten-seemann: I added some documentation.

templates/.github/workflows/go-check.yml
mvdan: I get wanting a separator, but I think the chances of confusion with the different separators are higher. Another option could be something like `.github/actions/setup/go-check`. If we don't want extra directories, personally I'd just stick with `go-check-setup`, but I also don't have strong feelings :)

Stebalien (approved):
I thought this was going to be a lot more complicated. This is nice.

templates/.github/workflows/go-check.yml
marten-seemann: Ok, I changed it to `-`.

marten-seemann:
> I thought this was going to be a lot more complicated. This is nice.

This was quite a lot of work to figure out. I'm really happy it turned out to be this simple in the end.

README.md
mvdan: :tada:

templates/.github/workflows/go-check.yml
mvdan: :pray:

mvdan (approved):
:runner:

PR #173, opened 2021-08-14
run go generate if enabled in the config

### LINKED ISSUE #123: consider running go generate
Should we run `go generate` in CI (probably in `go-check`)?
In general, there's no reason why generated code should not be up to date, _assuming_ that `go generate` is deterministic.

Not sure if that's a valid assumption. For quic-go it isn't, as we [generate](https://github.com/lucas-clemente/quic-go/blob/master/fuzzing/header/cmd/corpus.go) the fuzzing corpus there with a random input, but maybe that's a bug in quic-go that should be fixed?

I'm raising this issue because I just created a [PR](https://github.com/multiformats/go-multicodec/pull/49) in the multicodec repo that adds a separate workflow for that.

marten-seemann:
Fixes #123.

This PR introduces a new concept: Repositories can create a config file `.github/workflows/go-check-config.json` to enable an additional step in the `go-check` workflow: running `go generate ./...` to make sure that no files are modified or added.
This is a backwards-compatible change: If the config file doesn't exist, we don't run that step.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,8 @@ By storing them in a central place (here), and distributing them in an automated
 
 ## Customization
 
+### Additional Setup Steps
+
 Most repositories won't need any customization, and the workflows defined here will just work fine.
 Some repositories may require some pre-setup steps to be run before tests (or code checks) can be run. Setup steps for `go-test` are defined in `.github/actions/go-test-setup/action.yml`, and setup steps for `go-check` are defined in `.github/actions/go-check-setup/action.yml`, in the following format:
 
@@ -24,6 +26,21 @@ runs:
 ```
 
 These setup steps are run after the repository has been checked out and after Go has been installed, but before any tests or checks are run.
+
+### Configuration
+
+`go-check` contains an optional step that checks that running `go generate` doesn't change any files.
+This is useful to make sure that the generated code stays in sync.
+
+This check will be run in repositories that set `gogenerate` to `true` in `.github/workflows/go-check-config.json`:
+```json
+{
+  "gogenerate": true
+}
+```
+
+Note that depending on the code generators used, it might be necessary to [install those first](#additional-setup-steps).
+The generators must also be deterministic, to prevent CI from getting different results each time.
 
 ## Technical Details
 
--- a/templates/.github/workflows/go-check.yml
+++ b/templates/.github/workflows/go-check.yml
@@ -5,6 +5,8 @@ jobs:
   unit:
     runs-on: ubuntu-latest
     name: All
+    env:
+      RUNGOGENERATE: false
     steps:
       - uses: actions/checkout@v2
         with:
@@ -15,6 +17,12 @@ jobs:
       - name: Run repo-specific setup
         uses: ./.github/actions/go-check-setup
         if: hashFiles('./.github/actions/go-check-setup') != ''
+      - name: Read config
+        if: hashFiles('./.github/workflows/go-check-config.json') != ''
+        run: |
+          if jq -re .gogenerate ./.github/workflows/go-check-config.json; then
+            echo "RUNGOGENERATE=true" >> $GITHUB_ENV
+          fi
       - name: Install staticcheck
         run: go install honnef.co/go/tools/cmd/staticcheck@434f5f3816b358fe468fa83dcba62d794e7fe04b # 2021.1 (v0.2.0)
       - name: Check that go.mod is tidy
@@ -47,4 +55,17 @@ jobs:
           run: |
             set -o pipefail
             staticcheck ./... | sed -e 's@\(.*\)\.go@./\1.go@g'
+      - name: go generate
+        uses: protocol/multiple-go-modules@v1.2
+        if: (success() || failure()) && env.RUNGOGENERATE == 'true'
+        with:
+          run: |
+            git clean -fd # make sure there aren't untracked files / directories
+            go generate ./...
+            gitstatus=$(git status --porcelain)
+            if [[ -n "$gitstatus" ]]; then
+              echo "go generated caused changes to the repository:"
+              echo "$gitstatus"
+              exit 1
+            fi
 
--- a/templates/.github/workflows/go-check.yml
+++ b/templates/.github/workflows/go-check.yml
@@ -62,10 +62,10 @@ jobs:
           run: |
             git clean -fd # make sure there aren't untracked files / directories
             go generate ./...
-            gitstatus=$(git status --porcelain)
-            if [[ -n "$gitstatus" ]]; then
+            # check if go generate modified or added any files
+            if ! $(git add . && git diff-index HEAD --exit-code --quiet); then
               echo "go generated caused changes to the repository:"
-              echo "$gitstatus"
+              git status --short
               exit 1
             fi
 


templates/.github/workflows/go-check.yml
```diff
@@ -16,6 +18,12 @@ jobs:
       - name: Run repo-specific setup
         uses: ./.github/actions/go-check-setup
         if: hashFiles('./.github/actions/go-check-setup') != ''
+      - name: Read config
+        if: hashFiles('./.github/workflows/go-check-config.json') != ''
```
Stebalien, line 22: So, we _could_ just deploy a `test.env` file when we deploy CI.

Stebalien: That would let us drop jq.

templates/.github/workflows/go-check.yml
```diff
@@ -48,4 +56,19 @@ jobs:
           run: |
             set -o pipefail
             staticcheck ./... | sed -e 's@\(.*\)\.go@./\1.go@g'
+      - name: go generate
+        uses: protocol/multiple-go-modules@v1.2
+        if: ${{ env.RUNGOGENERATE == 'true' }}
```
Stebalien, line 61: If we only set this variable when true, I _think_ we could check `${{ env.RUNGOGENERATE }}` and drop the `== true`.

templates/.github/workflows/go-check.yml
```diff
@@ -5,6 +5,8 @@ jobs:
   unit:
     runs-on: ubuntu-latest
     name: All
+    env:
+      RUNGOGENERATE: false
```
Stebalien, line 9: If we drop our env variables in a json file, we can use `env: ${{ fromJson(...) }}` (I think).

templates/.github/workflows/go-check.yml
marten-seemann: Would that be easier? For a simple configuration like the one introduced here maybe, but in the future we might have more complex options.

templates/.github/workflows/go-check.yml
marten-seemann: Unfortunately that doesn't work. We can omit the `${{` though.

templates/.github/workflows/go-check.yml
marten-seemann: I'd like to avoid having to put JSON into the `test.env` file at some point.

Stebalien: I assume it would come down to running the logic on deploy, versus afterwards. But it's not critical either way.

templates/.github/workflows/go-check.yml
Stebalien: Oh well.

templates/.github/workflows/go-check.yml
```diff
@@ -16,6 +18,12 @@ jobs:
       - name: Run repo-specific setup
         uses: ./.github/actions/go-check-setup
         if: hashFiles('./.github/actions/go-check-setup') != ''
+      - name: Read config
+        if: hashFiles('./.github/workflows/go-check-config.json') != ''
+        run: |
+          if [[ $(jq -r .gogenerate ./.github/workflows/go-check-config.json) == "true" ]]; then
```
Stebalien, line 24: nit: use the `-e` flag in jq instead of parsing the output? (

Stebalien (approved):

templates/.github/workflows/go-check.yml
marten-seemann: Cool, I wasn't aware of that flag!

templates/.github/workflows/go-check.yml
```diff
@@ -48,4 +56,17 @@ jobs:
...
             set -o pipefail
             staticcheck ./... | sed -e 's@\(.*\)\.go@./\1.go@g'
+      - name: go generate
+        uses: protocol/multiple-go-modules@v1.2
+        if: env.RUNGOGENERATE == 'true'
+        with:
+          run: |
+            git clean -fd # make sure there aren't untracked files / directories
+            go generate ./...
+            gitstatus=$(git status --porcelain)
```
marten-seemann, line 66: Please pay attention to these lines when reviewing. Is this sane?

Stebalien: Ah, I see. We can't use `git diff --exit-code` because of untracked files?

This seems sane, but I'd also consider `git add . && git diff --exit-code`.

templates/.github/workflows/go-check.yml
```diff
@@ -48,4 +56,17 @@ jobs:
...
+            if [[ "$gitstatus" != "" ]]; then
```
Stebalien, line 67: `if [[ -z "$gitstatus" ]]`

Stebalien (approved):

templates/.github/workflows/go-check.yml
marten-seemann: 👍🏻

marten-seemann: Actually, this has to be `-n`, not `-z`.

templates/.github/workflows/go-check.yml
mvdan: From stackoverflow, I found `git ls-files --exclude-standard --others`, which I think I'd prefer slightly as it has a narrower purpose. If that doesn't work, I'd prefer Steven's suggestion with add+diff over just status, since it seems more intuitive.

README.md
```diff
@@ -25,6 +27,20 @@ runs:
 
 These setup steps are run after the repository has been checked out and after Go has been installed, but before any tests or checks are run.
 
+### Configuration
+
+`go-check` contains an optional step that checks that running `go generate` doesn't change any files.
+This is useful to make sure that the generated code stays in sync, however, it requires code generators to be deterministic.
```
mvdan, line 33: nit: I wouldn't say "however", but rather something a bit more positive like "please note that"

README.md
```diff
@@ -25,6 +27,20 @@ runs:
...
+
+This check will be run in repositories that set `gogenerate` to `true` in `.github/workflows/go-check-config.json`:
+```json
+{
+  "gogenerate": true
+}
+```
+
+Note that depending on the code generators used, it might be necessary to [install those first](#additional-setup-steps).
```
mvdan, line 42: Maybe you can put the determinism note here instead, as a second sentence:

> The generators must also be deterministic, to prevent CI from getting different results each time.

templates/.github/workflows/go-check.yml
mvdan: Wait, I'm confused now. Exactly what files are you checking for, here? Unstaged changes, *excluding* untracked and ignored files? Then my suggestion above might be wrong. I got it from https://stackoverflow.com/a/2659808/4931501, where there are other ideas depending on what you want.

Would also be good to document this inline, because I think I already misunderstood the code once :)

mvdan (approved):

README.md
marten-seemann: Thanks. Applied.

templates/.github/workflows/go-check.yml
marten-seemann: There are two things that should fail the build:
1. When `go generate` makes a change to a tracked file. Obviously, the PR should have updated that file.
2. When `go generate` adds a new file. The PR probably forgot to commit the generated file.

Ignored files shouldn't be considered. They're ignored for a reason.

If I understand correctly, `git ls-files --exclude-standard --others` does 2., but not 1.
`git add . && git diff --exit-code` doesn't work, since as soon as a file is `git add`ed, it doesn't show up in `git diff` any more.

marten-seemann: As far as I can tell, `git add . && git diff-index HEAD` would do 1. and 2. I guess that's nicer than `git status --porcelain`.

marten-seemann: Even better, it allows using the `--exit-code` flag, so we don't have to do string comparisons.

mvdan: FYI there is also diff --cached, aka --staged.

mvdan (approved):

PR #174, opened 2021-08-14
only run the automerge step if the PR was authored by web3-bot

marten-seemann:
Logically, this doesn't change anything. `automerge` depends on `automerge-check`, and `automerge-check` only runs when the PR was authored by web3-bot.
It does however prevent this job from being scheduled and then immediately being skipped.


mvdan (approved):
maybe document that inline? commit messages are great, but I'm not going to be reading all the git history as I try to understand the yaml :)

mvdan (approved):

PR #178, opened 2021-08-17
update to Go 1.17, update all repos with the updated workflows

marten-seemann:


mvdan (approved):
SGTM. We probably want to update staticcheck to https://github.com/dominikh/go-tools/releases/tag/2021.1.1 too, for its 1.17 support.

PR #184, opened 2021-08-19
Add reminder in README to add permissions to web3-bot

adlrocha:
I keep asking myself when setting up CI in a new repo *"why is the CI not being integrated?"* and it always ends up being that `web3-bot` doesn't have access to the repo. I suggest adding a quick note in the README for folks like me :)


galargh:
The usage section now includes this information: https://github.com/protocol/.github#usage

PR #186, opened 2021-08-20
run go mod tidy right after bumping the Go version number

marten-seemann:
Otherwise the workflow might fail, as happened here: https://github.com/protocol/.github/runs/3381383509?check_suite_focus=true:

image


marten-seemann:
I'm not sure if this PR is actually necessary. `go-libp2p-autonat/test` needs to be cleaned up first anyway.

PR #199, opened 2021-10-07
replace the $default-branch placeholder in workflows

marten-seemann:
This is #169, but now with the correct merge target (`master`).


mvdan:
>  Also, this change shouldn't trigger any new deployment. I asked for another review so you can double-check that it won't.

All pushes to master trigger copy workflows, no? It's just that some of them result in empty diffs so no PR is posted.

All this PR adds is the sed for the default-branch pattern, which as far as I can see is not used anywhere yet, so... I agree this should not cause any downstream PRs.

PR #200, opened 2021-10-10
make it possible to deploy versioning workflows to selected repositories

marten-seemann:
Repositories that set `deploy_versioning: true` in `config.json` will receive two new workflows: `releaser.yml` and `tagpush.yml`. To test these new workflows, we'll enable them for ~10 libp2p repositories first, before rolling them out to the entire org.

These files contain minimal workflow definitions, and call the corresponding workflows in this repo (to be added in #198 and #205). That way, we'll be able to update the actual workflow without deploying to all O(150) repos.

### DIFF
--- a/.github/workflows/copy-workflow.yml
+++ b/.github/workflows/copy-workflow.yml
@@ -29,7 +29,6 @@ jobs:
       NEEDS_UPDATE: 0
       INITIAL_TEST_DEPLOYMENT: 0
       GO_VERSION_BUMP: 0
-      HIGHEST_VERSION: ""
       FILES: ""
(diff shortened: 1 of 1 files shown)


mvdan:
> Repositories that set `deploy_versioning: true` in `config.json` will receive two new workflows:

Could we not base this off of a file being present in the target repo, e.g. a `VERSION.txt` file?

marten-seemann:
>  I imagine we'd remove these deploy_versioning bootstrap bits once all those repos have version.json in master.

We probably will remove `deploy_versioning` once we've tested out the new workflow in a few repositories, and run these steps unconditionally.

PR #202, opened 2021-10-12
fix example for OS-specific additional setup steps

marten-seemann:


marten-seemann:
It looks like GitHub (finally!) fixed this issue: https://github.com/actions/runner/issues/834#issuecomment-9567794.

PR #203, opened 2021-10-12
add a document describing our new versioning workflows

marten-seemann:
This PR adds a document explaining why we're moving away from releasing Go packages manually, and explains the workflows we're using instead and how to use them.

### DIFF
--- a/.github/workflows/tagpush.yml
+++ b/.github/workflows/tagpush.yml
@@ -7,7 +7,7 @@ env:
   ISSUE_BODY: |
     @{{ env.PUSHER }} just pushed a release tag: {{ env.TAGNAME }}.
     Please manually verify validity (using [\`gorelease\`](https://pkg.go.dev/golang.org/x/exp/cmd/gorelease)), and update \`version.json\` to reflect the manually released version, if necessary.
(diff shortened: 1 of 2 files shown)


marten-seemann:
> 1. "Every Go repository contains a version.json" file in the root directory may imply this is rolled out everywhere. This isn't the case though right? It's opt in, right? Assuming it's opt in, can we describe how to opt in to get these workflows.

That's correct for now. We'll test it for a week or two with a few selected repos, and then roll it out to every repo. Explicitly, this time we won't need repositories to sign up for this, we'll just deploy it to all repositories once we've collected a bit of experience with the new workflows.

> 2. In "https://github.com/protocol/.github/blob/master/.github/workflows/releaser.yml", I don't see us do a check that that the version number in version.json is valid for semver (before we do the push). I also didn't see that in https://github.com/protocol/.github/tree/master/.github/workflows . Do we have that as an option to ensure the right thing happens? If not, should we maybe comment or explain why?

That, among many other checks, is part of the release check workflow: https://github.com/protocol/.github/pull/205. This workflow is run when a PR is created that changes `version.json`.

PR #205, opened 2021-10-13
add a release check workflow

marten-seemann:
This workflow posts helpful information in a comment on PRs that modify `version.json`. Currently, this collects the output of `gorelease`, `gocompat` and the `git diff` of `go.mod` files.

Special care has to be taken to determine the correct version number to compare the new version to. Specifically, we cannot just compare to the highest released version, because the new release might be a patch release on an older release branch.

Example: Imagine a module that has releases for `v0.1.0`, `v0.2.0` and `v0.3.0`. When trying to cut a release `v0.2.1`, we need to base our comparisons on `v0.2.0`. When trying to cut a release `v0.3.1` or `v0.4.0`, we need to base our comparisons on `v0.3.0`.
This is currently achieved by using https://github.com/marten-seemann/semver-highest. I expect to refactor this tool based on @mvdan's feedback, and move it to the `protocol` org after that. Until then, this doesn't need to block us from moving forward with this PR though.

Sample output in: https://github.com/marten-seemann-test/target/pull/169

### DIFF
--- a/.github/workflows/release-check.yml
+++ b/.github/workflows/release-check.yml
@@ -30,7 +30,7 @@ jobs:
         # When trying to cut a release v0.3.1 or v0.4.0, we need to base our comparisons on v0.3.0.
         run: |
           git fetch origin --tags
-          go install github.com/marten-seemann/semver-highest@latest
+          go install github.com/marten-seemann/semver-highest@fcdc98f8820ff0e6613c1bee071c096febd98dbf
           v=$(semver-highest -target ${{ env.VERSION }} -versions $(git tag | paste -sd , -))
           echo "COMPARETO=$v" >> $GITHUB_ENV
       - name: Post output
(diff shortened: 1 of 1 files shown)


marten-seemann:
> Out of curiosity, why use both gorelease and gocompat? Do both of them bring features that we want which the other doesn't have?

Unfortunately, they do. For example, `gorelease` refuses to deal with releases that are not based on the most release, for example:
``` ~/src/go/src/github.com/libp2p/go-libp2p-quic-transport | @a4e84207#v0.12.0
❯ gorelease -base=v0.12.0
# summary
Cannot suggest a release version.
Can only suggest a release version when compared against the most recent version of this major: v0.13.0.

marten-seemann:
I suggest we keep both `gorelease` and `gocompat` for now, and if we discover that we can remove one of them, we can still do this later.

PR #210, opened 2021-10-16
include cross-package coverage in codecov

mvdan:
(see commit message)

### DIFF
--- a/templates/.github/workflows/go-test.yml
+++ b/templates/.github/workflows/go-test.yml
@@ -29,7 +29,10 @@ jobs:
       - name: Run tests
         uses: protocol/multiple-go-modules@v1.2
         with:
-          run: go test -v -coverprofile module-coverage.txt ./...
+          # Use -coverpkg=./..., so that we include cross-package coverage.
+          # If package ./A imports ./B, and ./A's tests also cover ./B,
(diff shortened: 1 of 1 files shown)


marten-seemann:
LGTM, but the merge target should be next, not master.

mvdan:
Ah, I did commit this on top of master, but my "post PR" alias just assumes I'm always targetting the default branch :)

mvdan:
PTAL

mvdan:
That's what I meant with the "ideal" part of the commit message: ideally, each package would have enough tests to cover itself well, say by at least 90%.

In practice however, in go-car/v2 we have the blockstore package that uses the index package. The blockstore package has very extensive tests, which already exercise >70% of the code in the index package. So, in my opinion, it's just wrong for codecov to say that the index package has near-zero test coverage.

It's also confusing to the user if they are told "this line in the index package is not covered", when if you add a panic, it makes `go test ./...` blow up.

I think either the current or new way of accumulating coverage are reasonable, depending on your expectations and how the information is used. If you prefer higher test quality, you probably want to not include cross-package coverage - but I think many PL repositories like go-car or storetheindex will take the practical approach of prioritizing integration tests to cover more code at once. And, with the way codecov says "line is not covered", I think including cross-package coverage is the better choice to avoid confusing the user :)

PR #215, opened 2021-10-18
improve readability of release-check workflow by renaming env vars

marten-seemann:
No functional change expected. This is literally just a "search-and-replace" change.


galargh:
Closing because the env vars that were being renamed here do not exist anymore.

PR #221, opened 2021-11-12
use msys2 on Windows

marten-seemann:
This is how our workflows would look like once we merge https://github.com/protocol/multiple-go-modules/pull/8.

### DIFF
--- a/templates/.github/workflows/go-test.yml
+++ b/templates/.github/workflows/go-test.yml
@@ -6,12 +6,15 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ "ubuntu", "windows", "macos" ]
         go: [ "1.16.x", "1.17.x" ]
+        os:
+          - { name: "ubuntu", shell: "bash" }
+          - { name: "macos", shell: "bash" }
+          - { name: "windows", shell: "msys2 {0}" }
     env:
       COVERAGES: ""
-    runs-on: ${{ matrix.os }}-latest
-    name: ${{ matrix.os}} (go ${{ matrix.go }})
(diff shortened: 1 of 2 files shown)


marten-seemann:
I updated the PR to use [v1.3 of multiple-go-modules](https://github.com/protocol/multiple-go-modules/releases/tag/v1.3).

templates/.github/workflows/go-test.yml
```diff
@@ -6,12 +6,15 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ "ubuntu", "windows", "macos" ]
         go: [ "1.16.x", "1.17.x" ]
+        os:
+          - { name: "ubuntu", shell: "bash" }
+          - { name: "macos", shell: "bash" }
+          - { name: "windows", shell: "msys2 {0}" }
```
mvdan, line 13: I wonder if we could derive the shell by default from the OS name, to reduce verbosity.

mvdan (approved):

templates/.github/workflows/go-test.yml
marten-seemann: Not without a separate step I think.

galargh: NOTE: this is a breaking change if someone started using `os.name` in their custom `go-test-setup/action.yml`. I don't expect there are many (if any) places like this so I can just keep an eye out for failing sync PRs once we propagate this change throughout the repos and fix where needed.

galargh:
Closing this since the alternative https://github.com/protocol/.github/pull/232 was merged into `next` branch.

PR #230, opened 2021-11-21
make it possible to cut release on non-default branches

### LINKED ISSUE #229: Cutting releases on a release branch
We only run the `releaser` workflow on the default branch: https://github.com/protocol/.github/blob/9c84fce2b8c4386e4e7558830825298c207e74ec/templates/.github/workflows/releaser.yml#L1-L5

That means it's not possible to cut a release on a release branch. I'm not sure what the right solution is. We definitely don't want to cut the release on every `push` to any branch, this would make it impossible to create a PR in the first place.

cc @mvdan @galargh

marten-seemann:
Fixes #229.

Implementing the solution suggested in https://github.com/protocol/.github/issues/229#issuecomment-974815879.

### DIFF
--- a/.github/workflows/releaser.yml
+++ b/.github/workflows/releaser.yml
@@ -6,11 +6,34 @@ jobs:
     runs-on: ubuntu-latest
     env:
       VERSION: ""
+      CREATETAG: "false"
+      BRANCH: ""
     steps:
       - uses: actions/checkout@v2
       - name: Determine version
         run: echo "VERSION=$(jq -r .version version.json)" >> $GITHUB_ENV
+      - name: Determine branch
+        run: echo "BRANCH=refs/heads/${{ github.event.repository.default_branch }}" >> $GITHUB_ENV
+      - name: Create a release, if we're on the default branch
+        run: echo "CREATETAG=true" >> $GITHUB_ENV
+        if: env.BRANCH == github.ref
+      - name: Determine if this is commit is a merged PR (if we're not on a non-default branch)
+        if: env.BRANCH != github.ref
+        id: getmergedpr
+        uses: actions-ecosystem/action-get-merged-pull-request@59afe90821bb0b555082ce8ff1e36b03f91553d9
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+      - name: Check if the "release" label was set on the PR
+        if: steps.getmergedpr.outputs.number != '' && env.BRANCH != github.ref
+        run: |
+          while IFS= read -r label; do
+            if [[ "$label" == "release" ]]; then
+              echo "CREATETAG=true" >> $GITHUB_ENV
+              break
+            fi
+          done <<< "${{ steps.getmergedpr.outputs.labels }}"
       - name: Create release
+        if: env.CREATETAG == 'true'
         run: |
           git fetch origin --tags
           if ! $(git rev-list ${{ env.VERSION}}.. &> /dev/null); then
--- a/.github/workflows/releaser.yml
+++ b/.github/workflows/releaser.yml
@@ -7,24 +7,24 @@ jobs:
     env:
       VERSION: ""
       CREATETAG: "false"
-      BRANCH: ""
+      DEFAULT_BRANCH: ""
     steps:
       - uses: actions/checkout@v2
       - name: Determine version
         run: echo "VERSION=$(jq -r .version version.json)" >> $GITHUB_ENV
       - name: Determine branch
-        run: echo "BRANCH=refs/heads/${{ github.event.repository.default_branch }}" >> $GITHUB_ENV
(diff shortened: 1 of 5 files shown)


templates/.github/workflows/releaser.yml
```diff
@@ -2,7 +2,6 @@ name: Releaser
 on:
   push:
     paths: [ 'version.json' ]
-    branches: [ $default-branch ]
```
marten-seemann, line 5: As far as I can tell, this should be safe to roll out: Once we merge this PR (to `master`), repos will start using the new `releaser` workflow, even if they haven't merged the PR updating the template yet. This means that `releaser` is only run for PRs merged into the default branch (which is the current behavior).

.github/workflows/releaser.yml
```diff
@@ -6,11 +6,33 @@ jobs:
...
     steps:
       - uses: actions/checkout@v2
       - name: Determine version
         run: echo "VERSION=$(jq -r .version version.json)" >> $GITHUB_ENV
+      - name: Determine branch
+        run: echo "BRANCH=refs/heads/${{ github.event.repository.default_branch }}" >> $GITHUB_ENV
+      - name: Create a release, if we're on the default branch
+        run: echo "CREATETAG=true" >> $GITHUB_ENV
+        if: env.BRANCH == github.ref
+      - name: Determine if this is commit is a merged PR (if we're not on a non-default branch)
```
galargh, line 20:
```suggestion
      - name: Determine if this commit is a merged PR (if we're not on a default branch)
```

galargh (approved):

.github/workflows/releaser.yml
galargh: `if we're not on a default branch` because `env.BRANCH != github.ref`, right?

Should we rename `env.BRNACH` to `env.DEFAULT_BRANCH` to make it clearer?

marten-seemann: Yes, looking at this now with a fresh mind `env.BRANCH` seems like exceptionally poor naming. Thanks for catching this!

marten-seemann: Fixed.

VERSIONING.md
```diff
@@ -34,6 +34,10 @@ The [release check workflow](.github/workflows/release-check.yml) will comment o
 
 As soon as the PR is merged into the default branch, the [releaser workflow](.github/workflows/releaser.yml) is run. This workflow cuts a new release on CI and pushes the tag.
 
+### Using a Release Branch
+
+Sometimes it's necessary to cut releases on a release branch. If you open Pull Request targeting a different branch than the default branch, a new release will only be created if the PR has the `release` tag.
```
mvdan, line 39:
```suggestion
Sometimes it's necessary to cut releases on a release branch. If you open a Pull Request targeting a branch other than the default, a new release will only be created if the PR has the `release` label.
```

.github/workflows/releaser.yml
```diff
@@ -6,11 +6,33 @@ jobs:
...
+        id: getmergedpr
+        uses: actions-ecosystem/action-get-merged-pull-request@59afe90821bb0b555082ce8ff1e36b03f91553d9
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+      - name: Check if the "release" label was set on the PR
+        if: steps.getmergedpr.outputs.number != '' && env.DEFAULT_BRANCH != github.ref
+        run: |
+          while IFS= read -r label ; do
+            if [[ "$label" == "release" ]]; then
+              echo "CREATETAG=true" >> $GITHUB_ENV
```
mvdan, line 31: break too?

.github/workflows/releaser.yml
```diff
@@ -6,11 +6,33 @@ jobs:
...
+      - name: Determine if this commit is a merged PR (if we're not on a default branch)
+        if: env.DEFAULT_BRANCH != github.ref
+        id: getmergedpr
+        uses: actions-ecosystem/action-get-merged-pull-request@59afe90821bb0b555082ce8ff1e36b03f91553d9
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+      - name: Check if the "release" label was set on the PR
+        if: steps.getmergedpr.outputs.number != '' && env.DEFAULT_BRANCH != github.ref
+        run: |
+          while IFS= read -r label ; do
```
mvdan, line 29:
```suggestion
          while IFS= read -r label; do
```

mvdan (approved):

VERSIONING.md
marten-seemann: Fixed during rebase.

.github/workflows/releaser.yml
marten-seemann: Fixed during rebase.

.github/workflows/releaser.yml
marten-seemann: Fixed.

PR #232, opened 2021-11-22
alternative: use msys2 on Windows

galargh:
This is an alternative way of using msys2bash on windows machines during go-test runs.

It prepends `C:/msys64/usr/bin` to `PATH`. After this, whenever we request `shell: bash`, `msys2`'s version will be picked ahead of `git`'s. 

With the original PR (https://github.com/protocol/.github/pull/221) the workflow wasn't able to find the paths specified in https://github.com/protocol/multiple-go-modules/blob/master/action.yml#L28:
```
New-Item : Cannot find path 'D:\a\_temp\setup-msys2\msys2.cmd' because it does not exist.
At line:1 char:1
+ New-Item -ItemType SymbolicLink -Path D:/a/_temp/setup-msys2/bash-or- ...
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : ObjectNotFound: (D:\a\_temp\setup-msys2\msys2.cmd:String) [New-Item], ItemNotFoundExcept 
   ion
    + FullyQualifiedErrorId : PathNotFound,Microsoft.PowerShell.Commands.NewItemCommand
 
Error: Process completed with exit code 1.
```

This alternative setup has been tested in https://github.com/libp2p/go-openssl/pull/19 (NOTE: the red checkmarks are due to 32 bit tests failing which is unrelated to this change). If you inspect the windows tests run please have a look at shell description - `shell: C:/msys64/usr/bin\bash.EXE`.

I've found this way of accomplishing the task by inspecting installed software on windows runners: https://github.com/actions/virtual-environments/blob/main/images/win/Windows2019-Readme.md#msys2

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,11 +21,8 @@ runs:
       run: echo "do some initial setup"
     - name: Step 2
       shell: bash
-      run: |
-        # note that it's not possible to use if: on the step, see https://github.com/actions/runner/issues/834
-        if [[ "${{ matrix.os }}" == "ubuntu" ]]; then
-          echo "do some Linux-specific setup"
-        fi
+      run: echo "do some Linux-specific setup"
+      if: ${{ matrix.os == 'ubuntu' }}
 ```
 
 These setup steps are run after the repository has been checked out and after Go has been installed, but before any tests or checks are run.
--- a/config.json
+++ b/config.json
@@ -35,6 +35,7 @@
   { "target": "ipfs/go-ds-pebble" },
   { "target": "ipfs/go-ds-redis" },
   { "target": "ipfs/go-ds-s3" },
+  { "target": "ipfs/go-ds-sql" },
   { "target": "ipfs/go-fetcher", "deploy_versioning": true },
   { "target": "ipfs/go-filestore", "deploy_versioning": true },
   { "target": "ipfs/go-fs-lock" },
--- a/config.json
+++ b/config.json
@@ -87,7 +87,7 @@
   { "target": "ipfs/iptb" },
   { "target": "ipfs/iptb-plugins" },
   { "target": "ipfs/pinbot-irc" },
-  { "target": "ipfs/tar-utils" },
+  { "target": "ipfs/tar-utils", "deploy_versioning": true },
   { "target": "ipld/codec-fixtures" },
   { "target": "ipld/go-car", "deploy_versioning": true },
   { "target": "ipld/go-codec-dagpb" },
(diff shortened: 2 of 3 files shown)


marten-seemann:
How does this work? Is the msys2 executable named `bash`?

galargh:
> How does this work? Is the msys2 executable named bash?

Yes, exactly! These are shells that are available on `windows-latest` machines:

### Shells ([source](https://github.com/actions/virtual-environments/blob/main/images/win/Windows2019-Readme.md#shells))
| Name          | Target                            |
| ------------- | --------------------------------- |
| gitbash.exe   | C:\Program Files\Git\bin\bash.exe |
| msys2bash.cmd | C:\msys64\usr\bin\bash.exe        |
| wslbash.exe   | C:\Windows\System32\bash.exe      |

By default, when we request `shell: bash` `C:\Program Files\Git\bin\bash.exe` is used. When we add `C:\msys64\usr\bin` to `PATH` (it is not on the path at all to begin with) then `C:\msys64\usr\bin\bash.exe` preceeds the `gitbash` version so it is picked for subsequent steps instead.

marten-seemann (commented):
That's interesting. This means that we basically set `bash = msys2` once, and all future steps will uses `msys2` instead of `bash`.
This also needs that we won't need the changes we made in [multiple-go-modules v1.3](https://github.com/protocol/multiple-go-modules/releases/tag/v1.3).

@mvdan, can you have a look at this?

mvdan (approved):
Certainly looks easier :) I'd say link to the docs in the YAML, so that if they change those paths in the future, at least we know where to look immediately.

galargh:
Thank you for the reviews! 

If I understand correctly, merging this will automatically propagate the changes across all repos (https://github.com/protocol/.github#technical-details). So I just wanted to make sure - do we have any additional procedures around merging such changes or should I just merge and check back if there are any unmerged web3-bot PRs a bit later?

marten-seemann:
> If I understand correctly, merging this will automatically propagate the changes across all repos (https://github.com/protocol/.github#technical-details). So I just wanted to make sure - do we have any additional procedures around merging such changes or should I just merge and check back if there are any unmerged web3-bot PRs a bit later?

Yes, so please don't merge to `master`. We try to keep the noise level low, so we just do a few deployments per year (ideally).
You can target the `next` branch. This branch has all the changes we want to get deployed next time we decide to deploy to all repos. Can you re-target this PR?

Issue #235, opened 2021-12-03
Setting up unified CI for js repos


achingbrain:
It would be great to get similar syncing of the `.github/workflows` folder for js repos as maintaining them is tedious work.

There are a few different setups with different CI requirements:

1. monorepos
2. repos that only target node.js
3. repos that target node.js and the browser

I spotted https://github.com/protocol/.github/tree/master/templates/.github/workflows - can we parameterise the repos in `config.json` to copy from subdirectories under `templates`?  Something simple like:

```js
[{
  "target": "filecoin-project/go-amt-ipld",
  "type": "go"
}, {
  "target": "libp2/libp2p-tcp",
  "type": "js/node"
}, {
  "target": "ipfs/js-datastore-s3",
  "type": "js/isomorphic"
}, {
  "target": "ipfs/js-unixfs",
  "type": "js/monorepo"
}
// etc
```

Then we can have `templates/go`, `templates/js/monorepo`, `templates/js/node`, `templates/js/isomorphic`, etc.

What do you think?  Some other way?

Also, some repos have unique build requirements, like `js-ipfs` testing itself with various repos from the https://github.com/ipfs-examples org - I guess maybe the copying of the templates could just overwrite files that are in the templates and leave others alone?  Then `js-ipfs` could have a `.github/workflows/examples.yml` file that would survive an update to `templates/js/monorepo` in this repo.

marten-seemann:
> Then we can have templates/go, templates/js/monorepo, templates/js/node, templates/js/isomorphic, etc.

That's an interesting idea. Are we reasonable sure that there won't be a lot of overlap (we could check in symlinks in git afaik, but that's probably not the nicest design. Alternatively, we could also have a separate section in the config mapping the `type` to an `array[file]`.

Another alternative would be to have multiple config files (`config-go.json`, config-js-monorepo.json`, ...).

We will also have to rework the copy workflow a bit. It has accumulated a bunch of Go-specific stuff. We'd probably need to refactor those steps into a separate file, and include that file depending on the `type`.

achingbrain:
I guess there would be a bit of overlap - the isomorphic CI would be a superset of the node CI, for example, though it's not a huge amount of code.

We could just have a js-monorepo build and a js-non-monorepo build - they'd invoke npm scripts with the `--if-present` flag - `test:node`, `test:browser` and so on.  Repos can opt-in to these jobs by declaring npm scripts with these names.  We have npm scripts like this everywhere already.

> multiple config files

[...]

PR #239, opened 2021-12-08
Prepare copy-workflow for setting up unified CI for JS repos

galargh:
## Description

Both `dispath.yml` and `copy-workflow.yml` have accumulated a fair share of `go` specific setup. In this PR I'm trying to generalise both of these workflows to enable [setting up unified CI for JS repos](https://github.com/protocol/.github/issues/235).

### Moving config JSONs to configs directory

With JS adoption we're expecting more JSON config files to be needed. I propose to store all the config JSON files in a designated `configs` directory so that they're easy to find. This also allows us to use matching patterns like `configs/*.json` in various places.

### Specifying what files should be copied for particular configuration

I have decided to create `defaults` object with `files` array field inside each config JSON file. The `files` array defines what files should be copied for that particular workflow by default. 

In dispatch workflow, each object from `repositories` array is mapped to the output of merging the `defaults` object with the repository object. Thanks to this, it is also possible to override the list of files to be copied on a per repository basis (or other fields that `defaults` might define):
```
{
  "defaults": { "files": ["a", "b", "c"] },
  "repositories": [
    { "target": "1", "description": "copy files a, b and c to repository 1" },
    { "target": "2", "files": ["a", "b"], "description": "copy only files a and b to repository 2" }
  ]
}
```

Having the list of files to be copied in the same file as the list of repositories makes it easier for repo owners to reason about the outcome of adding a repository to the repositories list.

### Handling language specific setup in copy-workflow

I have moved `go` specific setup to `copy-workflow-go` action. The action is performed iff the target object passed to the copy-workflow has `deploy_go` field set to `true`. I have used this pattern to provide specific setup for targets that request `deploy_versioning` too. 

It will also be applicable when we want to introduce more languages, like JS for example. We could just introduce new flags like `deploy_monorepo_js`, `deploy_isomorphic_js`, etc. 

## Testing

- [x] successfully ran `dispatch.yml` workflow on `testing` branch: https://github.com/protocol/.github/actions/runs/1553727642
- [x] successfully ran `copy-workflow.yml` workflow on `testing` branch: https://github.com/protocol/.github/actions/runs/1553730245
- [x] got https://github.com/protocol/.github-test-target/pull/33 automerged
- [x] changed the order of repositories in `configs/go.json` and failed to commit those changes because `pre-commit` hook prevented it
- [x] PR checks


.github/workflows/copy-workflow.yml
```diff
@@ -62,109 +58,54 @@ jobs:
...
+    - name: is initial workflow deployment
+      run: |
+        INITIAL_WORKFLOW_DEPLOYMENT=1
+        for f in $(jq -r '.[]' <<< $FILES); do
+          if [[ -f $TARGET_REPO_DIR/$f ]]; then
+            INITIAL_WORKFLOW_DEPLOYMENT=0
+            break
+          fi
+        done
+        echo "INITIAL_WORKFLOW_DEPLOYMENT=$INITIAL_WORKFLOW_DEPLOYMENT" >> $GITHUB_ENV
```
galargh, line 75: `INITIAL_WORKFLOW_DEPLOYMENT` is going to be `1` iff none of the files in the target repository exist yet.

.github/workflows/copy-workflow.yml
```diff
@@ -62,109 +58,54 @@ jobs:
...
-        if ! git diff --quiet; then
-          git add .
-          git commit -m "run gofmt -s"
-        fi
-    - name: determine files to add
-      run: echo "FILES=${{ toJson(github.event.inputs.files) }}" >> $GITHUB_ENV
-    - name: add more versioning workflows files, if necessary
+    - name: Run steps specific to go
+      if: matrix.cfg.deploy_go
+      uses: ./template-repo/.github/actions/copy-workflow-go
```
galargh, line 102: I'm not sure if there's a way to use `env.TEMPLATE_REPO_DIR` here.

.github/workflows/dispatch.yml
```diff
@@ -7,45 +7,59 @@ on:
...
+          TARGETS=()
+          for config in configs/*.json; do
+            if [[ $GITHUB_REF == refs/heads/testing && $config != configs/testing.json ]]; then
+              continue
+            fi
+            if [[ $GITHUB_REF != refs/heads/testing && $config == configs/testing.json ]]; then
+              continue
+            fi
+            DEFAULTS=$(jq -c '.defaults' $config)
+            TARGETS+=($(jq -c ".repositories[] | $DEFAULTS + ." $config))
```
galargh, line 36: If a specific `key` is present in the target object, it will override the one from defaults.

galargh:
Since this PR doesn't not modify any workflows in the target repositories I propose merging it into the `master` branch.

###### Testing after the rebase
- [x] dispatch workflow from `testing` branch: https://github.com/protocol/.github/actions/runs/1582674326
- [x] deploy workflow from `testing` branch: https://github.com/protocol/.github/actions/runs/1582676014
- [x] sync PR in .github-test-target: https://github.com/protocol/.github-test-target/pull/34

.githooks/pre-commit
```diff
@@ -1,12 +1,15 @@
...
-  echo "$output"
-  status=1
-fi
-rm $tmp
+for config in configs/*.json; do
+  tmp=$(mktemp)
+  git show :$config > $tmp
+  status=0
+  if ! output=$(.github/workflows/check-config.sh $tmp); then
+    echo "$tmp contains git show :$config output."
```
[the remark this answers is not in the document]
marten-seemann: There’s something wrong with that log string. `$tmp` will be a temporary file, so printing that file name doesn’t help the user. I’m also not sure what the `git show` is doing there (is it supposed to be a command?).

.github/workflows/check-config.yml
```diff
@@ -12,5 +12,8 @@ jobs:
       targets: ${{ steps.set-matrix.outputs.targets }}
     steps:
       - uses: actions/checkout@v2
-      - name: check if config.json is sorted alphabetically
-        run: .github/workflows/check-config.sh config.json
+      - name: check if configs/*.json is sorted alphabetically
```
marten-seemann, line 15:
```suggestion
      - name: check if config files are sorted alphabetically
```

.github/workflows/check-config.yml
```diff
@@ -12,5 +12,8 @@ jobs:
...
+        run: |
+          for config in configs/*.json; do
+            .github/workflows/check-config.sh $config
```
marten-seemann, line 18: Output the file name here, or do it the fancy way by using a [group](https://docs.github.com/en/actions/learn-github-actions/workflow-commands-for-github-actions#grouping-log-lines).

.github/workflows/copy-workflow.yml
marten-seemann: Can you add this as a comment?

.github/workflows/copy-workflow.yml
marten-seemann: That would be nicer, but maybe GH Actions doesn’t allow it?
Have you tried `${{ env.TEMPATE_REPO_DIR }}`.

.github/workflows/copy-workflow.yml
```diff
@@ -63,105 +55,56 @@ jobs:
...
+    - name: Run steps specific to versioning
+      if: matrix.cfg.deploy_versioning
+      uses: ./template-repo/.github/actions/copy-workflow-versioning
     - name: Add files
       run: |
-        for f in $(jq -r ".[]" <<< ${{ toJson(env.FILES) }}); do
+        for f in $(jq -r '.[]' <<< $FILES); do
```
marten-seemann, line 107: Just for the log output it would be nicer if we could continue using `env.FILES` (then GH Actions does the replacement before running the script).

.github/workflows/dispatch.yml
```diff
@@ -12,43 +12,53 @@ env:
...
       - uses: actions/checkout@v2
       - id: set-matrix
         run: |
-          CONFIG=config.json
-          if [[ $GITHUB_REF == refs/heads/testing ]]; then
-            CONFIG=config-testing.json
-          fi
-          TARGETS=$(jq '. | _nwise(${{ env.MAX_REPOS_PER_WORKFLOW }})' $CONFIG | jq -sc '. | to_entries')
-          echo "::set-output name=targets::$TARGETS"
+          TARGETS=()
```
marten-seemann, line 26: Why is this capitalized? Is it an env? Otherwise, it would probably be nicer to not capitalize this var.

configs/README.md
```diff
@@ -0,0 +1,35 @@
...
+| testing | repositories used for testing unified CI workflows |
+
+## Adding new repository to existing config file
+
+To add a new repository to an existing config file, add a new JSON object to the `repositories` key in the file.
+
+_**IMPORTANT**: Please remember to keep `repositories` sorted alphabetically by `target` value._
+
+*JSON object example:*
+```
```
marten-seemann, line 17:
```suggestion
```json
```

configs/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+After the PR with your change is merged, a copy workflow that runs in this repository will copy files(as defined by the `defaults.files` key of the config file or an overriden `files` key of your newly added object) to your repository.
+
+## Adding new config file
+
+When adding a new JSON config file, please follow the structure of other config files. In particular, `defaults.files` array and `repositories` array are required fields.
+
+*JSON config file example:*
+```
```
marten-seemann, line 28:
```suggestion
```json
```

configs/README.md
```diff
@@ -0,0 +1,35 @@
...
+{
+    "defaults": { "files": [] },
+    "repositories": []
+}
+```
+
+To customise the copy workflow further, you can add more fields to the `defaults` object. See `deploy_versioning` or `deploy_go` in [go.json](go.json#L8-L9) and how they are used in [copy-workflow.yml](../.github/workflows/copy-workflow.yml#L100-L105) for example.
```
marten-seemann, line 35: `deploy_versioning` is now gone.

configs/go.json
```diff
@@ -0,0 +1,192 @@
...
+    { "target": "multiformats/go-multiaddr-dns" },
+    { "target": "multiformats/go-multiaddr-fmt" },
+    { "target": "multiformats/go-multibase" },
+    { "target": "multiformats/go-multicodec" },
+    { "target": "multiformats/go-multihash" },
+    { "target": "multiformats/go-multistream" },
+    { "target": "multiformats/go-varint" },
+    { "target": "multiformats/ma-pipe" }
+  ]
+}
```
marten-seemann, line 192: Missing new line at the end of the file.

configs/testing.json
```diff
@@ -0,0 +1,8 @@
+{
+  "defaults": {
+    "files": []
```
marten-seemann, line 3: Don’t we need to deploy some file for testing?

configs/go.json
```diff
@@ -0,0 +1,192 @@
...
+    "files": [
+      ".github/workflows/automerge.yml", 
+      ".github/workflows/go-test.yml", 
+      ".github/workflows/go-check.yml", 
+      ".github/workflows/releaser.yml", 
+      ".github/workflows/release-check.yml", 
+      ".github/workflows/tagpush.yml"
+    ],
+    "deploy_versioning": true,
+    "deploy_go": true
```
marten-seemann, line 12: Why is this needed? Isn’t this obvious from the file name?

.github/workflows/stale-repos.yml
```diff
@@ -11,18 +11,19 @@ jobs:
...
-            if ! $exists; then
-              echo "Repository $repo does not exist."
-              status=1
-              continue
-            fi
-            if [[ $(echo "$output" | jq ".archived") == "true" ]]; then
-              echo "Repository $repo is archived."
-              status=1
-            fi
+          for config in configs/*.json; do
```
marten-seemann, line 14: Nit: Also use a group here to log the name of the config file.

.github/workflows/dispatch.yml
```diff
@@ -12,43 +12,53 @@ env:
...
       matrix:
         # We end up with multiple "dispatch" jobs,
-        # one per TARGETS "key" chunk above with a "value" array.
+        # one per BATCHES "key" chunk above with a "value" array.
         # For each "dispatch" job, matrix.cfg.value is an array, like:
         #
-        #   [{"target": "repo1", "target": "repo2"}]
+        #   [
+        #     {"target": "repo1", "files": [".github/workflows/go-check.yml"]}, 
+        #     {"target": "repo2", "files": [".github/workflows/go-check.yml", ".github/workflows/go-test.yml"]}
```
marten-seemann, line 51: Nice one!!!
I love that we can overwrite the list of files now!

.github/workflows/dispatch.yml
marten-seemann: Add something along these lines as a comment?

.githooks/pre-commit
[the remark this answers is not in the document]
galargh: In the original version the output was something like:
```
/var/folders/xx/k_hd05b922jds2mb4k9nzlv00000gn/T/tmp.wzfs3tsNhy contains git show:configs/go.json output.
Targets in /var/folders/xx/k_hd05b922jds2mb4k9nzlv00000gn/T/tmp.wzfs3tsNhy not sorted alphabetically:
...
```

I changed it to just display `Targets in configs/go.json not sorted alphabetically` as suggested.

.github/workflows/copy-workflow.yml
galargh: I added a comment about it. I've also tested the version using env once again in case I did some stupid mistake earlier but GH Actions still doesn't like it: https://github.com/protocol/.github/actions/runs/1583371137

.github/workflows/check-config.yml
galargh: I did use group, thanks for pointing me to it :)

.github/workflows/copy-workflow.yml
galargh: I restored the version that renders the values in the output.

configs/testing.json
galargh: I didn't want to include anything go specific in the testing configuration. My thinking was that it can be edited as needed. I added a section about it to the `configs/README.md` -> https://github.com/protocol/.github/blob/a71f87a0d506dfde49763755e91f1eb8732828cb/configs/README.md#testing

configs/README.md
galargh: Not entirely. I've left this part of the `deploy_versioning` flag - https://github.com/protocol/.github/blob/a71f87a0d506dfde49763755e91f1eb8732828cb/.github/actions/copy-workflow-versioning/action.yml

Other options here are:
1. move the steps from `copy-workflow-versioning` to https://github.com/protocol/.github/blob/a71f87a0d506dfde49763755e91f1eb8732828cb/.github/actions/copy-workflow-go/action.yml
2. execute `copy-workflow-versioning` only if ```".github/workflows/releaser.yml", ".github/workflows/release-check.yml"``` are being copied

Personally, I'd pick between what's currently implemented in this PR(explicit `deploy_versioning` flag) or option 2. Either of them would allow to opt-out or opt-in to versioning from other configurations. I wouldn't want to go for 1 because versioning setup, as it is now, is not go specific and if we could keep it this way it'd be great.

configs/go.json
galargh: Yes, we could. I set it up this way so other configurations can also opt-in to flagged sections of copy-workflow easily. For example, here's my `testing.json` file from `testing` branch -https://github.com/protocol/.github/commit/43476995428996a90ca95bf838f084ba1a710c68. As you can see, to mimic what `go.json` does, all I had to do was to copy the `defaults` from `go.json`.

If we wanted to remove the need for `deploy_go` flag in `go.json`, maybe we could add `source` field to the target object on which we operate in copy-workflow. Then we could change the if condition around the go section to `deploy_go == true || source == go.json`.

configs/README.md
marten-seemann: > I wouldn't want to go for 1 because versioning setup, as it is now, is not go specific and if we could keep it this way it'd be great.

That's true for `tagpush` and `release`, but not for `release-check`. That one is very much Go specific.
How do we deal with that?

configs/go.json
marten-seemann: Let's keep it the way it is. Having a separate `source` will probably be more confusing.

configs/README.md
galargh: Oh yeah, you're right about the `release-check`. Actually, now that I think about it, it's quite unlikely we'll be able to reuse `release` either. That's because both `js` and `rust`(which are the two most likely next candidates) have more complicated release processes(`npm`, `crate`) than just publishing git tags. So we'll have to create new workflows for them anyway.

All this makes me feel better about just moving the `copy-workflow-versioning` action content to `copy-workflow-go` action. Here's a PR for this if we want to go with it: https://github.com/protocol/.github/pull/256

One last tiny reservation I have is the ability to opt-out of versioning(which `go-ds-crdt` wants to do 😔). In the current implementation that would be possible by changing the repo object to:
```
{ 
  "target": "ipfs/go-ds-crdt", 
  "files":  [
      ".github/workflows/automerge.yml",
      ".github/workflows/go-test.yml",
      ".github/workflows/go-check.yml"
    ],
    "deploy_versioning": false
}
```

galargh: @marten-seemann What do you think? Personally, I think I still feel the best about the option that is currently implemented here.

marten-seemann: Not sure what to do with go-ds-crdt (and if we as an org are ok with a poorly tested repo like that), but given that they deactivated the workflow altogether, we probably shouldn't pay too much attention to them.

galargh: Which way do you think we should go here then? I'm happy either way.

.github/workflows/copy-workflow.yml
```diff
@@ -194,9 +139,9 @@ jobs:
         done
     - name: Check if we need to create a PR
       working-directory: ${{ env.TARGET_REPO_DIR }}
-      run: echo "NEEDS_UPDATE=$(git rev-list HEAD...origin/$(git rev-parse --abbrev-ref HEAD) --ignore-submodules --count)" >> $GITHUB_ENV
+      run: echo "NEEDS_UPDATE=$(git rev-list HEAD...origin/$(git rev-parse --abbrev-ref HEAD) --ignore-submodules --count 2> /dev/null || echo 1)" >> $GITHUB_ENV
```
marten-seemann, line 142: Nice one. This is a lot better than working with fake bools (strings).

marten-seemann (approved):
LGTM!

PR #240, opened 2021-12-09
deploy the versioning workflows to all repos

marten-seemann:
The versioning workflows will be presented at tonight's Demo Day. After that, we can use them everywhere.


.github/workflows/dispatch.yml
```diff
@@ -9,7 +9,7 @@ on:
 env:
   # We could use a higher number here. We use a small number just to make sure to create multiple batches.
   MAX_REPOS_PER_WORKFLOW: 16
-  FILES: '[ ".github/workflows/automerge.yml", ".github/workflows/go-test.yml", ".github/workflows/go-check.yml" ]' # a JSON array of the files to distribute
+  FILES: '[ ".github/workflows/automerge.yml", ".github/workflows/go-test.yml", ".github/workflows/go-check.yml", ".github/workflows/releaser.yml", ".github/workflows/release-check.yml", ".github/workflows/tagpush.yml" ]' # a JSON array of the files to distribute
```
mvdan, line 12: can we make this multi-line?

mvdan (approved):
How confident are you that this won't be a surprise to the majority of maintainers?

I'm thinking back to how it came as a surprise to Hannah for Graphsync, and I'm wondering what we're doing to make sure all maintainers are up to speed.

.github/workflows/dispatch.yml
galargh: We do have to refactor it soon to support JS anyway (here's my proposal - https://github.com/protocol/.github/pull/239) so I think we can leave it as is in this PR.

As for spreading the news about upcoming changes, I think Demo Day announcement is a great idea. Other thing I've seen working fairly well at other places was a dedicated announcement Slack channel, something like #w3dt-unified-ci-releases for example. We'd only ever post info about upcoming releases there and refrain from having other discussions (unless it's in a thread under post about the upcoming release). Do you think something like that could be a good fit at PL?

galargh (approved):

marten-seemann:
> How confident are you that this won't be a surprise to the majority of maintainers?

We can do another announcement in Slack before we hit the merge button.

PR #243, opened 2021-12-10
Replace concurrency setup with max-parallel=1 strategy in copy-workflow

galargh:
With current `concurrency` setup only the first and the last matrix job was run and the rest were cancelled with `Canceling since a higher priority waiting request for 'copy-*' exists`.

The [concurrency documentations](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#concurrency) explains this as follows:
```
Any previously pending job or workflow in the concurrency group will be canceled.
```

I think settings `strategy.max-parallel: 1` in `copy-workflow` might do what we want it to which is to run all the matrix jobs within one copy-workflow instance sequentially: https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_idstrategymax-parallel

One thing to watch out for is if the job sequence will continue after a failure but I think it will because of the `fail-fast: false` setting.

### DIFF
--- a/.github/workflows/copy-workflow.yml
+++ b/.github/workflows/copy-workflow.yml
@@ -22,6 +22,7 @@ jobs:
       fail-fast: false
       matrix:
         cfg: ${{ fromJson(github.event.inputs.targets) }}
+      max-parallel: 1
     env:
       TARGET_REPO_DIR: "target-repo"
       TEMPLATE_REPO_DIR: "template-repo"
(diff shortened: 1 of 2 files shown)


marten-seemann (approved):
According to my understanding:
* With ~180 repos, we'll end up with 9 copy workflows.
* In each of these copy workflows, only one job is running.
That means, we'll have a max of 9 concurrent deployments running.

galargh:
I've verified that `max-parallel: 1` + `fail-fast: false` work as we expect it to: https://github.com/protocol/cache-go-action/runs/4482926137?check_suite_focus=true

PR #244, opened 2021-12-10
try to avoid hitting secondary rate limit

galargh:
I propose to run a single batch at a time and introduce a 10s sleep after PR create to try and avoid hitting GitHub secondary rate limit. These values are quite caution because I'd want us to get a successful run across all repos and only then move on to trying to optimise it.

With these setting, the entire job should take around 180*30s=90m.

### DIFF
--- a/.github/workflows/dispatch.yml
+++ b/.github/workflows/dispatch.yml
@@ -7,11 +7,9 @@ on:
     branches: [ master, testing ]
 
 env:
-  # Number of repositories in a batch.
-  # Deployment jobs in the same batch are run sequentially.
-  # With ~180 repos, this means we'll run around 9 instances of the copy workflow in parallel.
-  # This is (hopefully) sufficient to prevent us from triggering GitHub Action's abuse detection mechanism.
-  MAX_REPOS_PER_WORKFLOW: 20
+  # The downstream workflow creates a matrix job per repository.
+  # 256 is the limit for number of matrix jobs a workflow can create.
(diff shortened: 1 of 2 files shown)


.github/workflows/dispatch.yml
```diff
@@ -7,11 +7,9 @@ on:
...
 
 env:
-  # Number of repositories in a batch.
-  # Deployment jobs in the same batch are run sequentially.
-  # With ~180 repos, this means we'll run around 9 instances of the copy workflow in parallel.
-  # This is (hopefully) sufficient to prevent us from triggering GitHub Action's abuse detection mechanism.
-  MAX_REPOS_PER_WORKFLOW: 20
+  # The downstream workflow creates a matrix job per repository.
+  # 256 is the limit for number of matrix jobs a workflow can create.
+  MAX_REPOS_PER_WORKFLOW: 256
```
marten-seemann, line 12: That would mean that we're just running one job in parallel. I'd hope that we can avoid reducing the number that far.

galargh: Yes, that was my intention. I'd want to see if we can get a fully successful run with a single job running at a time. Then, with the next release we could try to optimise it.

galargh: Since it's not going out today, let me put this PR back into draft for now.

galargh:
Closing this since I agree now that making all deploys last at least 90m is a not acceptable. I'm working on a better approach to fix this issue and I will open a new PR once it's ready.

Issue #228, opened 2021-11-19
Create GitHub releases in releaser/tag-pusher workflows


galargh:
The suggestion here is for the [versioning workflows](https://github.com/protocol/.github/blob/master/VERSIONING.md#using-the-versioning-workflows) and [manual version pushes](https://github.com/protocol/.github/blob/master/VERSIONING.md#dealing-with-manual-pushes) to create a GitHub release among with the version tag.

Below, I've gathered some ideas for the process from Slack discussions.

###### How to generate release notes?
- use [auto generated release notes](https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes); [API access to this feature](https://docs.github.com/en/rest/reference/repos#generate-release-notes-content-for-a-release)
- for versioning workflow - we could use the comments/description of the PR updating the version to create the release notes (Q: how to access PR in the workflow?)
- for manual push workflow - we could use the comments/description of the issue that was created as a result of pushing tag manually (Q: should we check if the release was created manually too?)

###### How to verify the GitHub release before subscribers get notified about it?
- initially create the GitHub release in a draft state (make it possible for owners to change this behaviour once they gain enough confidence in the workflow); [API access to this feature](https://docs.github.com/en/rest/reference/repos#create-a-release)

marten-seemann:
> for manual push workflow - we could use the comments/description of the issue that was created as a result of pushing tag manually (Q: should we check if the release was created manually too?)

Ideally, this workflow is never executed. It only exists because GitHub doesn't allow us to restrict tag pushes (yet).

> initially create the GitHub release in a draft state (make it possible for owners to change this behaviour once they gain enough confidence in the workflow); API access to this feature

Can we do this, even though the git commit (and the tag) doesn't exist yet? 

> for versioning workflow - we could use the comments/description of the PR updating the version to create the release notes (Q: how to access PR in the workflow?)

There are multiple things we can do:
1. Use [artifacts](https://docs.github.com/en/actions/advanced-guides/storing-workflow-data-as-artifacts) to store the text until the PR is merged, and then copy it from the artifact to the release notes. Problem: there might be multiple PRs suggesting the same release, and only one of them gets merged.
2. Is is possible to run a workflow when a PR is closed (and then check that it wasn't just closed, but actually merged), see https://github.community/t/trigger-workflow-only-on-pull-request-merge/17359. Importantly, this does _not_ give us access to the merge commit, so this workflow might run before the `releaser` workflow has actually created the tag.

[...]

PR #255, opened 2021-12-17
chore: remove go-ds-crdt

hsanjuan:
Sorry, the release flow that was enabled replaces a "git tag; git push" operation, which takes me 1 second, with a PR, a 40 minute build and maintenance of a file with redundant information. And it does not generate an automatic changelog based on commit messages or publish a github release as I expected so it literally takes 0 work from me. All this is just adding crust that costs me time. I will re-enable this project when this changes.

### DIFF
--- a/config.json
+++ b/config.json
@@ -28,7 +28,6 @@
   { "target": "ipfs/go-ds-badger" },
   { "target": "ipfs/go-ds-badger2" },
   { "target": "ipfs/go-ds-bitcask" },
-  { "target": "ipfs/go-ds-crdt" },
   { "target": "ipfs/go-ds-flatfs" },
   { "target": "ipfs/go-ds-leveldb" },
   { "target": "ipfs/go-ds-measure" },


Stebalien:
The goal of the release flow is to check the version number against semver. Basically, it catches breaking changes in patch releases.

Unfortunately, GitHub doesn't have any way to run these checks pre-release/tag. I agree that the new flow is really annoying, but if it's a 40 minute thing, CI is likely taking too long anyways.

marten-seemann:
> it does not generate an automatic changelog based on commit messages or publish a github release as I expected so it literally takes 0 work from me

That's actually something we're working on (see #228), and will be rolled out in the next iteration of the workflows.
As @Stebalien points out, these new workflows prevent us from messing up semver (as we've done countless times in the past!).

hsanjuan:
This is all very useful in general.

But since this was introduced I have lost a lot of time.  The CI runs 6 or 8 jobs and not being able to do it in parallel makes it 40 minutes total time. The CI wants to run all tests together and is unable to adapt to the project (separate datastore suite from local tests). At first it was mis-configured to the point tests failed because they took longer than expected. I lost a lot of time trying to explain how this is 0 help as I'm actively maintaining the project and it already has the tests it needs.

But I still left it around in the end. Now release-management has been introduced, something I never needed nor asked for. I have tried it and discovered it makes me lose more time and does not help me.

I am not arguing that these unified CI is unhelpful or not needed in general, I am trying to argue that it is getting my way, repeatedly, for this project in particular at this point in time. I babysit 3 or 4 other small repos where I don't mind it. I don't understand what is so hard to accept that this project, by what it is, by what it tests, by how it is maintained right now, may constitute an exception to the unified CI and perhaps it is not the end of the world if we do not blanket-roll it over it, specially after the maintainer requests it.

Can we please grant the exception?

marten-seemann (approved):
Fair point. Thank you for elaborating @hsanjuan, and sorry that this took so long.

Unified CI is supposed to make maintainers' life easier, not the other way around. If at any point in the future you feel like you want to reenable the workflows, feel free to open a PR here. @galargh has been working on decoupling the testing workflows from the versioning workflows, so very soon it will also be possible to just opt-in into one of them.

hsanjuan:
> so very soon it will also be possible to just opt-in into one of them.

Thank you, happy re-opting into the testing then.

PR #256, opened 2021-12-17
move deploy_versioning to deploy_go action

galargh:


marten-seemann:
Can you help me understand why you're moving this? This logic is independent from the language in use, isn't it?

galargh:
> Can you help me understand why you're moving this? This logic is independent from the language in use, isn't it?

I was just trying to visualise how getting rid of `deploy_versioning` could look like. As a follow-up to https://github.com/protocol/.github/pull/239#discussion_r771314652

I think I still like the version where we keep the `deploy_versioning` variable best.

Issue #254, opened 2021-12-16
release-check workflow fails when triggered from 3rd party repos


marten-seemann:
Example: https://github.com/libp2p/go-doh-resolver/runs/4548155702?check_suite_focus=true

It’s doing a git diff `${{ env.COMPARETO }}..${{ github.event.pull_request.head.sha }}`. The v0.3.1 tag exists in our as well as in the forked repo, the problem is probably the commit SHA.

@aschmahmann mentioned that we might need to do something like here: https://github.com/ipfs/go-ipfs/blob/master/.circleci/main.yml#L108-L124

marten-seemann:
> The issue is that if a workflow is initiated by a pull from a fork, then the created GITHUB_TOKEN can have at most read privileges to everything. 

Is a `GITHUB_TOKEN` with `read` permissions not allowed to post on a PR? That would be surprising to me, after all, a contributor with the same permissions is allowed to comment on issues and PRs.

galargh:
Yes, `GITHUB_TOKEN` gets only `read` permissions on pull requests: https://github.com/libp2p/go-doh-resolver/runs/4559777662?check_suite_focus=true#step:1:23 which makes the commenter action fail https://github.com/libp2p/go-doh-resolver/runs/4559777662?check_suite_focus=true#step:14:76 because of lacking permissions https://github.com/marocchino/sticky-pull-request-comment#error-resource-not-accessible-by-integration

I'm pretty sure what happens here is that the Pull Request permission cannot express ownership. So you're either granted access to writing to all PRs within a repo or none.

galargh:
Release check, when run on forks, now outputs the release notes which can be copied manually. If this proves to be a common occurrence, we can set up an cron job that is going to post such release notes - closing for now.

PR #257, opened 2021-12-17
make release-check pass on PRs from forked repositories

galargh:
This PR tries to address https://github.com/protocol/.github/issues/254

This PR makes sure that runs of `release-check` workflow succeed for PRs from forks BUT it does have to disable auto posting of the messages on such PRs. However, it makes the message easily discoverable/copyable from the workflow run. An idea for a future improvement would be to create something like `release-check-cron` that would loop over recently run `release-check`s in the repository and post comments to the PRs if they haven't been posted yet. That would be possible because a workflow that runs on cron schedule from master is allowed to have elevated privileges. 

###### Fixed issues
- replaced uses of `github.event.pull_request.head.sha` with `HEAD` since, on PRs from forked repos, the former points at a sha from the fork(which is not fetched) and the latter works as expected because it points at the local HEAD(which is the target branch fast forwarded to the head)
- replaced uses of `github.event.pull_request.head.ref` with `github.event.pull_request.head.label` because the latter also includes repository name in format of `repo_name:branch_name`(it works as expected when pasted at the and of github's compare URL)

###### Persisting issues
- workflows run from forked repos have at most read permissions on the autogenerated GitHub token(otherwise we'd have a serious security issue on our hand) so we are not able to post comment on the PR

###### Workarounds
- set notice on the `release-check` run that contains the message if running on a forked PR
- archive the message as an artifact of the `release-check` run if running on a forked PR

### Example:
**Release check run**: https://github.com/libp2p/go-doh-resolver/actions/runs/1592009640
**Message manually copied to a PR**: https://github.com/libp2p/go-doh-resolver/pull/17#issue-1083168982

### DIFF
--- a/.github/workflows/release-check.yml
+++ b/.github/workflows/release-check.yml
@@ -136,10 +136,10 @@ jobs:
       - name: Set a notice message on run
         run: |
           echo "$MESSAGE" > message.md
-          MESSAGE="${MESSAGE//'%'/'%25'}"
-          MESSAGE="${MESSAGE//$'\n'/'%0A'}"
-          MESSAGE="${MESSAGE//$'\r'/'%0D'}"
-          echo "::notice ::$MESSAGE"
+          message="${MESSAGE//'%'/'%25'}"
+          message="${message//$'\n'/'%0A'}"
+          message="${message//$'\r'/'%0D'}"
+          echo "::notice ::$message"
         if: env.INITIAL_RUN == 'false' && env.COMPARETO != '' && github.event.pull_request.head.repo.full_name != github.repository
       - name: Archive message
         uses: actions/upload-artifact@v2
--- a/.github/workflows/release-check.yml
+++ b/.github/workflows/release-check.yml
@@ -104,28 +104,39 @@ jobs:
           If you wish to cut a release once this PR is merged, please add the \`release\` label to this PR.
           EOF" >> $GITHUB_ENV
         if: github.base_ref != github.event.repository.default_branch
-      - name: Post output
-        uses: marocchino/sticky-pull-request-comment@82e7a0d3c51217201b3fedc4ddde6632e969a477 # v2.1.1
+      - run: |
+          echo 'MESSAGE<<EOF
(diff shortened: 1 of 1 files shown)


.github/workflows/release-check.yml
```diff
@@ -104,29 +104,47 @@ jobs:
...
-            ```
-            ${{ env.GOCOMPAT }}
-            ```
-            ${{ env.RELEASE_BRANCH_NOTE }}
-
+          message: ${{ env.MESSAGE }}
+      - name: Set a notice message on run
+        run: |
+          echo "$MESSAGE" > message.md
+          MESSAGE="${MESSAGE//'%'/'%25'}"
```
marten-seemann, line 139: Better: use a local variable here (`msg`?)

.github/workflows/release-check.yml
```diff
@@ -104,29 +104,47 @@ jobs:
...
+          MESSAGE="${MESSAGE//$'\n'/'%0A'}"
+          MESSAGE="${MESSAGE//$'\r'/'%0D'}"
+          echo "::notice ::$MESSAGE"
+        if: env.INITIAL_RUN == 'false' && env.COMPARETO != '' && github.event.pull_request.head.repo.full_name != github.repository
+      - name: Archive message
```
marten-seemann, line 144: This doesn't make sense without a workflow that reads from the artifacts, does it?

marten-seemann (approved):
That's very unfortunate, but I guess there's nothing we can do about it.
I suppose the case where external contributors cut releases is too rare to actually care about this too much, so I'd be fine with just posting this into the jobs output.

@mvdan, wdyt?

.github/workflows/release-check.yml
galargh: Yes, this would only be more convenient than a notice message if we were to, for example, create a cron job that would pick up such unposted release check messages and post them to their PRs.

If we don't think that's worth the effort, then let's clean it up: https://github.com/protocol/.github/pull/259

Issue #250, opened 2021-12-11
automerge should include workflow from this repo


marten-seemann:
Currently, we ship the `automerge` workflow to all repos. That's because back when we started Unified CI, it was not possible to include workflows from other repos.

Now that this is possible, we should use this feature (analogous to how we include the versioning workflows).

PR #260, opened 2021-12-21
Make automerge a reusable workflow

galargh:
Closes https://github.com/protocol/.github/issues/250

It makes the `automerge.yml` that we distribute to all the repos use `.github/workflows/automerge.yml` from this repo. It makes it easier for us to improve/fix the automerge logic by removing the need to update the file in all the repos.

###### Testing
- [x] https://github.com/protocol/.github-test-target/pull/36 got automerged using the new automerge setup


marten-seemann (approved):
LGTM.
Have we tested this? I'm a bit worried that GH Actions might not allow us to get the status of workflow runs from a workflow that's not defined in the same repo.

galargh:
> LGTM.
Have we tested this? I'm a bit worried that GH Actions might not allow us to get the status of workflow runs from a workflow that's not defined in the same repo.

I'm pretty sure I did test this but let me double-check since I didn't add any note about it in the description before.

Issue #36, opened 2021-03-29
Opening many PRs triggers GitHub's abuse detection mechanism


marten-seemann:
For example here: https://github.com/protocol/.github/actions/runs/698173097.

It looks like GitHub is imposing a limit on how many PRs we can create. I'm not exactly sure what the limit here is, but batch 1 went through, creating 12 PRs, and then about half the jobs (~16) in batch 0 succeeded. Therefore, the limit is probably somewhere between 25 and 30. I have no idea how long it takes for the window to open up again.
This is bad. This is not a lot.

One possible solution would be to build `sleep` statements into the `dispatch` workflow. Unfortunately, we'd have to assume that every run of the copy workflow makes changes to *all* files, if we want to reliably avoid workflow failures. If the workflow is managing 5 files (we're already managing 4 right now), each batch can only copy to 5 target repositories. After that, we'd have to sleep for a yet to be determined time before copying to the next 5 repos.

mvdan:
I've also been wondering if we're creating too much PR/commit noise. Look at https://github.com/ipld/go-ipld-adl-hamt/commits/master, for example; I get that this stuff is still experimental and changing a lot, but ten commits in the past month is still a lot. And I fear that rate will only increase over time as we make the CI setup more complete and customizable.

Have you thought about batching PRs? For example, aim to have at most one open "CI sync" PR per repository. If one already exists and is open, even if it updated a different file, just add more commits to it.

I get "one commit per PR" makes for easier reviews, but honestly, many of the changes being made in PRs are trivial to review anyway - pretty much just "yeah, the automation worked as expected". And reviewing+merging one medium-sized PR with automated changes is easier than doing that across half a dozen PRs.

marten-seemann:
That probably makes sense. It would mean that we can't use a matrix for the files we're updating any more, as we'd need one workflow run that creates a branch, and then adds commits to that branch one after the other. That's not necessarily a bad thing, but it does mean that we'll end up with more bash code in our workflow.

The interesting question is how this would work if such a PR doesn't get merged before the next time we make a change to the template. I _think_ that we'd just create another branch based on the most recent change, and that the "Create PR" action will basically do a force-push of that branch. That would actually be my preferred solution here, but I'd have to try that out to be sure.

Just noting that this won't completely solve our rate limiting problem though: We still can't open more than 25 PRs at the same time, but at least we'd be able to service 25 repositories at the same time (and that number would remain constant, even if at some point we ship 10 different files).
[...]

PR #261, opened 2021-12-22
open PRs sequentially but continue updating branches in parallel

galargh:
This PR tries to address https://github.com/protocol/.github/issues/36.

## Prerequisites

It includes the changes introduced in https://github.com/protocol/.github/pull/239 so this PR is going to be ready for review once that PR is merged.

## Description

When deploying changes to all the repositories, we're hitting a [secondary rate limit](https://docs.github.com/en/rest/overview/resources-in-the-rest-api#secondary-rate-limits) due to opening too many PRs too quickly. I believe the only problematic call involved in the process is the [Create a Pull Request](https://docs.github.com/en/rest/reference/pulls#create-a-pull-request) API call. The documentation mentions that hitting the endpoint too quickly may lead to hitting the limit.

That's why I propose to reorganise the `dispatch`/`copy-workflow` procedure in the following way:
1. Prepare batches of up to 256 targets.
1. For each batch(sequentially):
    1. Trigger `copy-workflow` with the batch; for each target(in up to 20 parallel jobs):
        1. Checkout the default branch
        1. Copy files as instructed by the config
        1. Check if there are any changes
        1. Push the changes to remote `web3-bot/sync` branch IF there are any changes
    1. Wait for the triggered `copy-workflow` to complete
    1. Find all the matrix jobs from the `copy-workflow` that pushed changes to remote `web3-bot/sync` branch
    1. For all the found matrix jobs(sequentially):
        1. Create a PR
        1. Sleep for 3 seconds

The main idea here is to do as much of parallelisable work inside `copy-workflow` and parallelise it as much as possible. That only leaves the PR creation calls to be executed sequentially. Furthermore, by executing PR creation calls conditionally(only if needed) we're able to keep the runtime when we only update a handful of repos small.

We could further optimise this procedure by performing the PR creation part in parallel to `copy-workflow` execution. I think that would be possible because the `copy-workflow` doesn't execute any GH API calls. However, I propose that we postpone any work in this direction until we deem it necessary because it would add even more complexity to the `dispatch` workflow.

### Removing PR create API call from `copy-workflow`

To accomplish this I remove the use of `peter-evans/create-pull-request` completely and replace it with:
```
git checkout -B web3-bot/sync
git push origin web3-bot/sync -f
```

These 2 commands ensure that our local state, after copying all the necessary files, is reflected in the remote `web3-bot/sync` branch. They are only executed if needed.

### Dispatching `copy-workflow` and waiting for the completion

Unfortunately, `benc-uk/workflow-dispatch` is not capable of waiting for the dispatched workflow completion. It can also be replaced by a simple `gh workflow run` call(`gh` is already installed on the runners) so I decided to do just that.

Sadly, the latter doesn't know how to wait for the workflow completion either. Nor does it return the dispatched workflow `id` so we have to be clever about acquiring it. I decided to:
1. record a timestamp before dispatching `copy-workflow`;
1. dispatch `copy-workflow`;
1. request the latest run of the `copy-workflow` until it's start date is after the recorded timestamp;
1. record the id of that run of the `copy-workflow`.

This works because `copy-workflow`s are being dispatched sequentially which means there is never more than one `copy-workflow` running at a time.

Once we have the id in hand, we can use `gh run watch` to wait for the run to complete.

### Finding all the repositories that need PRs created

Once again, the Actions framework doesn't make it easy to retrieve this information. It doesn't support job outputs from matrix jobs(or rather it does but it only persist the output of the latest job which is not useful here).

The workaround I came up with is to:
- inside `copy-workflow`: conditionally execute the branch push step;
- inside `dispatch`:
    1. find all the matrix jobs executed by the `copy-workflow` run;
    1. filter them by the status of the branch push step execution;
    1. map the remaining jobs to the job names which are in `{owner}/{repository}` format.

### Creating the PRs

It can be achieved with `gh api -X POST "/repos/$job_name/pulls"`.

I also added a 3 second sleep after the GH API call. I decided to choose 3 seconds because that value comes up in multiple places across GH API so there's a chance it's going to work here too.

## Overhead

I ran `dispatch` workflow in my account to measure the overhead of this setup. To achieve this I:
- replaced the GH API call to create a PR in `dispatch` with an echo,
- removed everything but the push step from `copy-workflow`,
- replaced the push step in `copy-workflow` with an echo,
- made the push step to always execute.

This means that the overhead numbers are for a case where we have to create PRs(and sleep for 3 seconds) for all the repositories in the configuration.

| Repositories | Repositories per batch | Max parallel factor | Time spent in `copy-workflow` | Time spent "creating PRs" | Total execution time |
| --- | --- | --- | --- | --- | --- |
| 176 | 100 | 10 | &nbsp;&nbsp;&nbsp;2m 53s<br/>+ 1m 51s<br/>= 4m 44s<br/> | &nbsp;&nbsp;&nbsp;5m 27s<br/>+ 4m 7s<br/>= 9m 34s | 15m 15s |

## Testing

- [x] measured overhead of the process: [dispatch](https://github.com/galargh/.github/actions/runs/1608507431) + [copy-workflow(batch 0)](https://github.com/galargh/.github/actions/runs/1608508926) + [copy-workflow(batch 1)](https://github.com/galargh/.github/actions/runs/1608537400)
- [x] deployed new workflow to `.github-test-target`: [dispatch](https://github.com/protocol/.github/actions/runs/1611673020) + [copy-workflow](https://github.com/protocol/.github/actions/runs/1611674889) + [PR](https://github.com/protocol/.github-test-target/pull/35)

### DIFF
--- a/.github/workflows/dispatch.yml
+++ b/.github/workflows/dispatch.yml
@@ -1,4 +1,4 @@
-# Trigger the execution of copy-workflow.yml, in batches of 256 repositories.
+# Trigger the execution of copy workflow in batches.
 # This workflow is needed since GitHub Actions limits the matrix size to 256 jobs.
 # We use one job per repository per batch.
 
@@ -8,13 +8,13 @@ on:
 
 env:
   # Number of repositories in a batch.
-  # Deployment jobs in the same batch are run sequentially.
+  # Matrix jobs within a copy workflow run are run in parallel.
   # 256 is the upper limit on the number of matrix jobs.
-  # Batching too many targets together can result in
+  # Batching too many repositories together can result in
   #  could not create workflow dispatch event: HTTP 422: inputs are too large.
-  #  when issueing dispatch_workflow event.
+  # This value should be higher than max-parallel in copy workflow.
   MAX_REPOS_PER_WORKFLOW: 100
-  # Number of seconds to wait before starting to watch workflow status.
+  # Number of seconds to wait before starting to watch copy workflow run.
   # Unfortunately, the interval on the watch is not configurable.
   # The delay helps us save on GH API requests.
   WORKFLOW_COMPLETION_CHECK_DELAY: 60
@@ -62,8 +62,9 @@ jobs:
         #     {"target": "repo2", "files": [".github/workflows/go-check.yml", ".github/workflows/go-test.yml"]}
         #   ]
         #
-        # The triggered copy-workflow jobs use that final array as their matrix.
-        # Since max-parallel here is 1, we'll end up with at most copy-workflow@max-parallel + 1 parallel jobs.
+        # The triggered copy workflow runs use that final array as their matrix.
+        # Since max-parallel here is 1, we'll end up with at most max-parallel from copy workflow + 1 parallel jobs.
+        # 20 is the upper limit on parallel jobs on a free plan.
         cfg: ${{ fromJson(needs.matrix.outputs.batches) }}
       max-parallel: 1
     env:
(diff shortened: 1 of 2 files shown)


galargh:
@marten-seemann Have you had a chance to look at this? I'm interested to hear what you think about this approach. I was thinking of trying it out on the next major release of unified CI for Go.

.github/workflows/dispatch.yml
```diff
@@ -51,18 +58,87 @@ jobs:
...
+          # delays checking copy workflow's run status to save on GH API requests
+          sleep ${{ env.WORKFLOW_COMPLETION_CHECK_DELAY }}
+
+          # checks every 3 seconds until the copy workflow run's status is completed
+          # redirects the stdout to /dev/null because it is very chatty
+          gh run watch "${{ steps.run.outputs.id }}" --repo "${{ env.WORKFLOW_REPO }}" > /dev/null
+      - id: jobs
+        name: Find copy workflow jobs that pushed changes
+        env:
+          STEP_NAME: 'Force push web3-bot/sync branch'
```
laurentsenta, line 108: nitpick:
in the copy-workflow, the name is `'Force push ${{ env.GITHUB_USER }}/sync branch'`,
so if you change the `GITHUB_USER` in one place, you break the workflow in another place.
Could we comment & warn about the coupling in copy-workflow?
Could we enforce a unique name? Like "Force push unified-ci branch"?

galargh: Good point, I made this change - https://github.com/protocol/.github/pull/261/commits/d9cfe7cab23dc3650e4925f93d4755103fbc9caa - so that it's not suggested that the username is configurable.

galargh:
Merging - as per verbal approval from @laurentsenta

Issue #57, opened 2021-04-06
Templating


Stebalien:
When trying to figure out how to deploy issue templates, I've realized that I need some level of templating. E.g.,

- The org's name.
- The org's security email.
- The repo's name.
- The repo's URL.

marten-seemann:
What a coincidence. I just talked with @andyschwab and we decided to create different users for every org. That way, if one user gets compromised, the attacker can't take over *all* PL repos.
This also means that we'll have to use some kind of templating for the `autorebase` workflow.

marten-seemann:
Simplest option: Use `%PLACEHOLDER%` in the yml template file, and then run `sed` in the copy workflow.

That doesn't scale very well though.

Do you have any suggestion for a more elegant solution?

Stebalien:
`m4` or maybe go templates?

PR #262, opened 2021-12-22
add template engine functionality to support generic templates

galargh:
## Template engine

I created a [README.md](https://github.com/protocol/.github/blob/486f183d83a80dd1a48589b8bfbf1672f8167064/templates/README.md) that describes the template engine from the end user perspective i.e. what the syntax is, what objects are available, etc.

On the backend, I use perl to:
1. perform a regex match on `\${{{(.?*)}}}`;
1. execute jq with the matched group as a filter and context JSON object;
1. replace the match with jq output

## Extra files

The newly added `extra_files` config field can be used to request copying of extra files while not modifying the `files` value from `defaults`.

## Use-case example

See https://github.com/protocol/.github/pull/434

## Testing

- [x] deployed new workflow to `.github-test-target` from [testing](https://github.com/protocol/.github/commit/51fdddf66cb4dd8f98a48e9ce78df8cae73374f3): [dispatch](https://github.com/protocol/.github/actions/runs/1612112471) + [copy-workflow](https://github.com/protocol/.github/actions/runs/1612114954) + [PR](https://github.com/protocol/.github-test-target/pull/36/commits/2d166c3148dff853543e5159dc4feb176c90abf7)


templates/README.md
```diff
@@ -0,0 +1,83 @@
...
+
+This directory contains template files that can be expanded and copied over to the configured repositories.
+
+## Header
+
+The [header](header.yml) is prepended to all the files before they are copied to the target repositories.
+
+## Contexts
+
+You can access context information during template expansion(before copy is performed).
```
marten-seemann, line 11:
```suggestion
You can access context information during template expansion (before copy is performed).
```

.github/workflows/copy-workflow.yml
```diff
@@ -106,13 +107,25 @@ jobs:
...
+        replacement="\$filter = \".\$1\"; \$filter =~ s/'/'\"'\"'/g; \`echo '${context//\'/\'\"\'\"\'}' | jq -jc '\$filter'\`"
         for f in $(jq -r '.[]' <<< ${{ toJson(env.FILES) }}); do
           echo -e "\nProcessing $f."
           # add DO NOT EDIT header
           tmp=$(mktemp)
           cat $TEMPLATE_REPO_DIR/$TEMPLATE_DIR/header.yml $TEMPLATE_REPO_DIR/$TEMPLATE_DIR/$f > $tmp
-          # replace $default-branch with this repo's GitHub default branch
-          sed -i "s:\$default-branch:${{ env.DEFAULTBRANCH }}:g" $tmp
+          # replace template contexts with values from the JSON context object
+          perl -pi -e "s#$regexp#$replacement#ge" $tmp
```
marten-seemann, line 128: https://xkcd.com/1171/

templates/README.md
```diff
@@ -0,0 +1,83 @@
...
+    "config": {
+        "example": {
+            "greeting": "Hello"
+        },
+        "files": [],
+        "extra_files": [".github/workflows/example.yml"],
+        "target": "protocol/.github-test-target"
+    },
+    "github": {
+        "default_branch": "master"
```
marten-seemann, line 64: Why is this a config value? This can be easily obtained from the github context.

marten-seemann (commented):
Is the `sync-release-assets` the only use case for this?
It seems like this could be easier accomplished by just publishing a GH Actions Workflow and manually calling this workflow from the two (?) repos that care about it, no?

templates/README.md
galargh: I don't think [github context](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context) has the name of the default branch. However, even if it did, we might want to use that information prior to the context being available, just as we did here for example: https://github.com/protocol/.github/blob/dfff3e56fa8ec5ecd0ce0188489e62f94a1be9fe/templates/.github/workflows/releaser.yml#L5

And it looks like the JS workflow will need it soon too: https://github.com/protocol/.github/blob/5cec3f5077e1d6ce52c17c73c85460793085feaf/templates/.github/workflows/js-test-and-release.yml#L5

galargh:
> Is the sync-release-assets the only use case for this?

It is a concrete example that I decided to use when working on templating functionality but it is not the only use case. Deploying issue templates is another example that was mentioned here https://github.com/protocol/.github/issues/57

> It seems like this could be easier accomplished by just publishing a GH Actions Workflow and manually calling this workflow from the two (?) repos that care about it, no?

Yes, it is possible to manually copy [the workflow file](https://github.com/protocol/.github/blob/406b18365f417643ce1aaf92916f3b2a49868f37/templates/.github/workflows/sync-release-assets.yml) to interested repositories to achieve the same but I believe it is a pretty good fit for this repository to be the one distributing it. At the moment there are only 2 but I think there might be more to come. @lidel might be able to elaborate on this.

templates/README.md
marten-seemann: Right, it's not available in `if` conditions, so it makes sense to _somehow_ replace this.
I'm wondering though if configuring it in a config file is the right way to go. The copy workflow now determines the name of the default branch on the fly, which makes for a smoother workflow when repos rename their default branch.

galargh: Oh yes, I do completely agree and that's why I didn't change it. The `github` template context object is still created on the fly during `copy-workflow`'s execution. See https://github.com/protocol/.github/blob/406b18365f417643ce1aaf92916f3b2a49868f37/.github/workflows/copy-workflow.yml#L110-L113

It's only the `config` template context object that comes from the "parsed JSON"/whatever the input to the `copy-workflow` was.

marten-seemann: What does setting the `github.default_branch` do then? I'm confused now :)

galargh: This discussion is happening under a line from an example of a [context container](https://github.com/protocol/.github/blob/486f183d83a80dd1a48589b8bfbf1672f8167064/templates/README.md#context-container) - an object that we can refer to during template expansion.

If I write a template that has `${{{ some.json.path }}}` somewhere in its content, then that string will be replaced with the output of `echo "$CONTEXT_CONTAINER" | jq -jc '.some.json.path'` - the value of `some.json.path` inside the context container.

The context container, at least for now, has 2 keys - `config` and `github`.

The `github`'s value is an object that is created on the fly during the execution of the `copy-workflow`. Currently, it has one key - `default_branch` - but we can add more in the future if we want to.

The `config`'s value is an object that was passed to `copy-workflow` as [the input](https://github.com/protocol/.github/blob/406b18365f417643ce1aaf92916f3b2a49868f37/.github/workflows/copy-workflow.yml#L21). So it's, for example, a merge of defaults https://github.com/protocol/.github/blob/406b18365f417643ce1aaf92916f3b2a49868f37/configs/go.json#L2-L13 and a specific repository config https://github.com/protocol/.github/blob/406b18365f417643ce1aaf92916f3b2a49868f37/configs/go.json#L15

marten-seemann: Got it, so the context container is generated automatically. Maybe that's worth spelling out explicitly in the README?

galargh: Good idea, let me add this :)

galargh: Done! I also added and example of JSON config and expanded template to the examples - https://github.com/protocol/.github/pull/262/commits/676e4905ddea2065df45b02b515a75d7179af45a.

galargh:
@lidel What do you think about using unified CI framework to distribute `sync-release-assets.yml`?

.github/workflows/copy-workflow.yml
```diff
@@ -101,13 +102,26 @@ jobs:
...
         for f in $(jq -r '.[]' <<< ${{ toJson(env.FILES) }}); do
           echo -e "\nProcessing $f."
           # add DO NOT EDIT header
           tmp=$(mktemp)
           cat $TEMPLATE_REPO_DIR/$TEMPLATE_DIR/header.yml $TEMPLATE_REPO_DIR/$TEMPLATE_DIR/$f > $tmp
-          # replace $default-branch with this repo's GitHub default branch
-          sed -i "s:\$default-branch:${{ env.DEFAULT_BRANCH }}:g" $tmp
+          # replace template contexts with values from the JSON context object
+          perl -pi -e "s#$regexp#$replacement#ge" $tmp
+          mv $tmp $TEMPLATE_REPO_DIR/$TEMPLATE_DIR/$f
```
laurentsenta, line 124: Is it worth investigating the case where a file name contains a space?

laurentsenta:
2 notes:
- I couldn't get an exit code != 0 if the templated variable was missing, could be my env when I tested, the expansion was "just" `null` without error.
- Do we have a complexity threshold when we want to move these to "something not shell scripts"?

galargh:
> * I couldn't get an exit code != 0 if the templated variable was missing, could be my env when I tested, the expansion was "just" `null` without error.

That's expected and that's fine for what it's trying to be right now.

> * Do we have a complexity threshold when we want to move these to "something other than shell scripts"?

I added a note for ourselves to consider alternatives if we ever have to develop the templating feature further. I think that's the threshold 😛

PR #271, opened 2022-01-07
check github actions yamls

galargh:


galargh:
I need to fix the action code first because it doesn't glob subdirectories properly.

PR #272, opened 2022-01-07
check github actions yamls

### LINKED ISSUE #76: broken yaml stopped workflow execution
image

Do we need a YAML linter? :(

galargh:
Resolves #76 

###### Description

This PR adds GitHub workflow that runs on PR in this repo which performs YAML linting. It gets the correct schema from https://www.schemastore.org/api/json/catalog.json (same as VS Code YAML extension).

I target `next` in this PR because I had to modify `go-test.yml` slightly to make the linter happy.

### DIFF
--- a/.github/workflows/check-yaml.yml
+++ b/.github/workflows/check-yaml.yml
@@ -1,10 +1,10 @@
 on:
   pull_request:
     paths:
-      - '.github/actions/*/*.yml'
-      - '.github/workflows/*.yml'
-      - 'templates/.github/actions/*/*.yml'
-      - 'templates/.github/workflows/*.yml'
+      - '**/action.yml'
+      - '**/action.yaml'
+      - '**/.github/workflows/*.yml'
+      - '**/.github/workflows/*.yaml'
 
 name: Check YAML
 
@@ -13,10 +13,11 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - uses: galargh/validate-yaml-schema@0be36a8865108d9fc0b2cae260dc0f9f649fac96 # forked v1.0.3
+      # TODO: switch to mainline once https://github.com/nwisbeta/validate-yaml-schema/pull/21 is merged
+      - uses: galargh/validate-yaml-schema@c8060d4a389d0eec986a90665bf0c496bf1f52e8
         with:
           yamlSchemasJson: |
             {
-              "https://json.schemastore.org/github-action.json": [".github/actions/*/*.yml", "templates/.github/actions/*/*.yml"],
-              "https://json.schemastore.org/github-workflow.json": [".github/workflows/*.yml", "templates/.github/workflows/*.yml"]
+              "https://json.schemastore.org/github-action.json": ["**/action.yml", "**/action.yaml"],
+              "https://json.schemastore.org/github-workflow.json": ["**/.github/workflows/*.yml", "**/.github/workflows/*.yaml"]
             }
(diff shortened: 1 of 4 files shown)


.github/workflows/check-yaml.yml
```diff
@@ -0,0 +1,22 @@
...
+      - 'templates/.github/workflows/*.yml'
+
+name: Check YAML
+
+jobs:
+  check-yaml:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: galargh/validate-yaml-schema@0be36a8865108d9fc0b2cae260dc0f9f649fac96 # forked v1.0.3
```
mvdan, line 16: why was this forked? if a change was necessary, perhaps we can send a PR and phase out the fork sooner than later.

templates/.github/workflows/go-test.yml
```diff
@@ -10,8 +10,8 @@ jobs:
         go: [ "1.16.x", "1.17.x" ]
     env:
       COVERAGES: ""
-    runs-on: ${{ matrix.os }}-latest
-    name: ${{ matrix.os}} (go ${{ matrix.go }})
+    runs-on: ${{ format('${0}-latest', matrix.os) }}
```
mvdan, line 13: was the old syntax wrong? I'm a bit confused because it apparently worked just fine :)

.github/workflows/check-yaml.yml
```diff
@@ -0,0 +1,22 @@
+on:
+  pull_request:
+    paths:
+      - '.github/actions/*/*.yml'
+      - '.github/workflows/*.yml'
+      - 'templates/.github/actions/*/*.yml'
+      - 'templates/.github/workflows/*.yml'
```
mvdan, line 7: How about `**/actions/**.yml` and `**/workflows/**.yml`? Seems more general and less likely to miss files as we reorganize or add directories.

mvdan (approved):

.github/workflows/check-yaml.yml
galargh: It's because it didn't work for YAMLs nested more than one level deep. Of course, I've already created a PR - https://github.com/nwisbeta/validate-yaml-schema/pull/20

templates/.github/workflows/go-test.yml
galargh: It worked but the schema from https://json.schemastore.org/github-workflow.json didn't account for it. I'll see if there's a way to express what we were doing in that schema - if so I'll create a PR but, personally, I'd be fine with the setup proposed here too.

.github/workflows/check-yaml.yml
mvdan: SGTM - add a TODO with the link perhaps, so we remember to update when merged

templates/.github/workflows/go-test.yml
mvdan: I see. Slight preference for keeping what we had, because it was simpler, but I also don't think it matters too much :)

.github/workflows/check-yaml.yml
galargh: I thought there was no point doing that because we had to be explicit in the action inputs anyway but it turns out that's not entirely true - https://github.com/nwisbeta/validate-yaml-schema/pull/21. Let me adjust it to `**/action.yml`, `**/.github/workflows/*.yml`.

templates/.github/workflows/go-test.yml
mvdan: SGTM! Thanks for being thorough :)

Issue #40, opened 2021-03-31
Script to label all issues/PRs matching some query


Stebalien:
We occasionally need to go through and "fix" issue labels. It would be nice to have a script or a script framework where we could do things like: set `need/triage` on all issues lacking a `kind/` label.

mvdan:
There are some actions in the github marketplace that do this sort of thing, but their configs tend to be rather stiff. I found one that applies PR labels based on files changed, or another that applies issue labels based on keywords.

I think a script would work fine here. You could even do it in Bash with a mix of https://github.com/cli/cli and `jq`, but another way would be to bite the bullet and use Go with either the REST or GraphQL client.

galargh:
I recently created actions library that contains one to [find issues or pull requests by query](https://github.com/protocol/github-api-action-library/tree/master/find-content-by-query). I use it to find all the content that matches certain queries and add it to a project: https://github.com/galargh/.github/blob/master/.github/workflows/add-project-items.yml

We could definitely reuse it for this use case too. The query to find all issues without [kind labels](https://github.com/ipfs/community/blob/e0812a37d7a611bbf1f3b37b0bd00c017f98da8f/ISSUE_LABELS.md#kind) in all of our org would look something like that:
```
is:issue -label:kind/architecture,kind/bug,kind/discussion,kind/enhancement,kind/maintenance,kind/support,kind/test org:protocol org:ipfs org:ipfs-shipyard org:libp2p org:ipld org:multiformats
```

If we also add the action that updates issues with new labels to the library then we'll have all the building blocks there.

PR #275, opened 2022-01-10
add a workflow that adds a label to content matching query

galargh:
Resolves https://github.com/protocol/.github/issues/40

###### Description

This PR adds a workflow that accepts a [issue or pull reuqest query](https://docs.github.com/en/search-github/searching-on-github/searching-issues-and-pull-requests) and a label and adds that label to all the issue or pull requests matching the query.

If label is left blank, the workflow will just print the repository name with the owner and the number of the issue or pull request it would have added a label to.

*NOTE*: It will only be able to add labels to issues/PRs in repositories `web3-bot` was invited to. 

###### Testing
- [x] https://github.com/galargh/.github/actions/runs/1681726803 added `wontfix` to https://github.com/protocol/w3dt-stewards/issues/4 and https://github.com/protocol/w3dt-stewards/issues/3 based on the following query `is:issue is:closed repo:protocol/w3dt-stewards test`

### DIFF
--- a/.github/workflows/add-label-by-query.yml
+++ b/.github/workflows/add-label-by-query.yml
@@ -17,7 +17,7 @@ jobs:
       GITHUB_TOKEN: ${{ secrets.WEB3BOT_GITHUB_TOKEN }}
     steps:
       - id: content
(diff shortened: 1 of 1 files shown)


laurentsenta (approved):
LGTM,
would it make sense to add a doc with a few examples?

PR #280, opened 2022-01-13
simplify Go version upgrade procedure

### LINKED ISSUE #111: define how we do Go version upgrades
We probably want to write this up, as we'll have to come back to this every 6 months.

Necessary steps:
- bump the Go version in `go.mod` (via `go mod edit`)
- remove Go version `N-2` and add version `N` in `go-test` and `go-check`
- run `gofmt`
- run `go fix`
- run `go mod tidy`

All but the first point pose a challenge: We **only** want to run them once, when doing the version upgrade, not on every deployment.

galargh:
Resolves #111

###### Description

This PR simplifies how Go versions are defined by making https://github.com/protocol/.github/blob/3104214787d953097a1d849fff039298e489f8a0/configs/go.json#L13-L15 the only place which explicitly states the version strings. All other places in which we referred to Go versions will transitively read from this config from now on.

It also describes all the steps that are required to bump Go version - i.e. update the version strings in the list. 

###### Testing

- [x] https://github.com/protocol/.github-test-target/pull/39

### DIFF
--- a/.github/actions/copy-workflow-go/action.yml
+++ b/.github/actions/copy-workflow-go/action.yml
@@ -4,23 +4,28 @@ description: Copy workflow steps specific to go
 runs:
   using: "composite"
   steps:
+    # GitHub Actions Expressions do not support last item access/array length retrieval
+    - id: go
+      run: echo "::set-output name=version::$(jq -r '.[-1]' <<< '${{ toJSON(matrix.cfg.go.versions) }}')"
+      shell: bash
     - uses: actions/setup-go@v3
       with:
         # This should be the same Go version we use in the go-check workflow.
         # go mod tidy, go vet, staticcheck and gofmt might behave differently depending on the version.
-        go-version: "1.19.x"
+        go-version: ${{ steps.go.outputs.version }}
     - name: bump go.mod go version if needed
       uses: protocol/multiple-go-modules@v1.2
       with:
         working-directory: ${{ env.TARGET_REPO_DIR }}
         run: |
           # We want our modules to support two Go versions at a time.
-          # As of August 2022, Go 1.19 is the latest stable.
           # go.mod's Go version declares the language version being used.
           # As such, it has to be the minimum of all Go versions supported.
-          # Bump this every six months, as new Go versions come out.
-          TARGET_VERSION=1.18
-          PREVIOUS_TARGET_VERSION=1.17
+          TARGET_VERSION='${{ matrix.cfg.go.versions[0] }}'
+          TARGET_VERSION="${TARGET_VERSION%.x}"
+          TARGET_MAJOR_VERSION="${TARGET_VERSION%.[0-9]*}"
+          TARGET_MINOR_VERSION="${TARGET_VERSION#[0-9]*.}"
+          PREVIOUS_TARGET_VERSION="$TARGET_MAJOR_VERSION.$(($TARGET_MINOR_VERSION-1))"
 
           # Note that the "<" comparison doesn't understand semver,
           # but it should be good enough for the foreseeable future.
--- a/.github/actions/copy-workflow-go/action.yml
+++ b/.github/actions/copy-workflow-go/action.yml
@@ -25,6 +25,10 @@ runs:
           TARGET_VERSION="${TARGET_VERSION%.x}"
           TARGET_MAJOR_VERSION="${TARGET_VERSION%.[0-9]*}"
           TARGET_MINOR_VERSION="${TARGET_VERSION#[0-9]*.}"
+          # Assumptions:
+          # - all versions are targetted incrementally
+          # - no versions are skipped
+          # - patch version is never pinned explicitly 
           PREVIOUS_TARGET_VERSION="$TARGET_MAJOR_VERSION.$(($TARGET_MINOR_VERSION-1))"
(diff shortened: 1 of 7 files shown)


templates/.github/workflows/go-check.yml
```diff
@@ -13,7 +13,7 @@ jobs:
           submodules: recursive
       - uses: actions/setup-go@v2
         with:
-          go-version: "1.17.x"
+          go-version: "${{ go.versions[-1] }}"
```
marten-seemann, line 16: Does this work? Won't it panic? ;)

templates/.github/workflows/go-test.yml
```diff
@@ -7,7 +7,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ "ubuntu", "windows", "macos" ]
-        go: [ "1.16.x", "1.17.x" ]
+        go: ${{ config.go.versions }}
```
marten-seemann, line 10: How is this being replaced? Wouldn't this need the templating PR?

marten-seemann (commented):
Not sure I understand how this works. Have we tested it?

templates/.github/workflows/go-check.yml
galargh: I forgot to stage my changes 🤦  Added them in a new commit now so that the comment history makes sense: https://github.com/protocol/.github/pull/280/commits/9bd61de7c19ceb2be6e0f13ea180c408330c280a

Now it reads:
```
go-version: ${{{ config.go.versions[-1] }}}
```

`[-1]` is a valid reference to the last element in `jq` which is what the templating engine I proposed uses. Sadly, it is not in GitHub Actions Expressions.

templates/.github/workflows/go-test.yml
galargh: Again, forgot to stage my latest changes - https://github.com/protocol/.github/pull/280/commits/9bd61de7c19ceb2be6e0f13ea180c408330c280a

You've spotted correctly that it should use the template engine syntax - `${{{ config.go.versions }}}`.

And yes - this PR requires #262. I mentioned it in the `Prerequisites` section of the description.

galargh:
> Have we tested it?

No, not yet. I tested some sections of it but not the entire thing. I was planning on doing that once the prerequisites are satisfied. That's also why I'm leaving this PR in `draft` mode for now.

.github/actions/copy-workflow-go/action.yml
```diff
@@ -4,23 +4,28 @@ description: Copy workflow steps specific to go
...
           # go.mod's Go version declares the language version being used.
           # As such, it has to be the minimum of all Go versions supported.
-          # Bump this every six months, as new Go versions come out.
-          TARGET_VERSION=1.18
-          PREVIOUS_TARGET_VERSION=1.17
+          TARGET_VERSION='${{ matrix.cfg.go.versions[0] }}'
+          TARGET_VERSION="${TARGET_VERSION%.x}"
+          TARGET_MAJOR_VERSION="${TARGET_VERSION%.[0-9]*}"
+          TARGET_MINOR_VERSION="${TARGET_VERSION#[0-9]*.}"
+          PREVIOUS_TARGET_VERSION="$TARGET_MAJOR_VERSION.$(($TARGET_MINOR_VERSION-1))"
```
laurentsenta, line 28: Is it correct we're making two assumptions here?

- The script assumes we're doing every migration one by one, we can't go from `1.41` to `1.43`.
- We never pin the patch version, it's always `.x`.

Should we document these?

galargh:
```suggestion
          # Assumptions:
          # - all versions are targetted incrementally
          # - no versions are skipped
          # - patch version is never pinned explicitly 
          PREVIOUS_TARGET_VERSION="$TARGET_MAJOR_VERSION.$(($TARGET_MINOR_VERSION-1))"
```

laurentsenta (approved):
(adding a note on the documentation, rest lgtm)

configs/README.md
```diff
@@ -38,3 +38,7 @@ To customise the copy workflow further, you can add more fields to the `defaults
 ## Testing
 
 You can use [testing](https://github.com/protocol/.github/tree/testing) branch for worklow/configuration testing. Once you push your changes to the branch, a [dispatch](../.github/workflows/dispatch.yml) workflow will be triggered. The workflow will use [testing.json](testing.json) configuration file only. You can manipalate that configuration file as needed(you can copy all the `defaults` from [go.json](go.json) for [example](https://github.com/protocol/.github/commit/43476995428996a90ca95bf838f084ba1a710c68)).
+
+## Upgrading Go
+
+To upgrade Go, modify the `defaults.go.versions` array in the [Go config](go.json). Keep the array sorted in an increasing order.
```
laurentsenta, line 44:
```suggestion
To upgrade Go, modify the `defaults.go.versions` array in the [Go config](go.json).

Remember to:
- Keep the array sorted in increasing order,
- Upgrade versions incrementally. Do not skip a version,
- never pin the patch version (`"1.19.x"` is correct, `"1.19.8"` is incorrect).
```

galargh: Great suggestion! Thanks :)

PR #281, opened 2022-01-13
always add a version.json file if it doesn't exist

### LINKED ISSUE #264: add a version.json file for repos that don't have any releases
Example: https://github.com/libp2p/go-libp2p-resource-manager/pull/2

We should probably add a `version.json` with an empty version string if there are no previous releases. That would make it more intuitive to cut the first release.

galargh:
Resolves #264

It will add a `version.json` file with the following content in repos that don't have releases yet:
```
{
  "version": ""
}
```

Do you think we should target `next` or `master` with this? If we target `master`, it will create new PRs but only in repos without `version.json` which shouldn't be too bad.


marten-seemann (approved):
No strong opinion which branch to target.
Given that the next deployment is just a month or so away, targeting next should be fine.

PR #286, opened 2022-01-17
allow specifying go vet flags

galargh:
This PR will make it possible to provide flags for `go vet` command via `go-check-config.json` object.

###### Why?

One of the reasons why https://github.com/ipld/go-ipld-prime-proto/pull/24 cannot be merged is because `go vet` fails with `composite literal uses unkeyed fields` (see https://github.com/ipld/go-ipld-prime-proto/runs/4483048466?check_suite_focus=true). Unfortunately, the solution of changing the code to avoid this error doesn't work here because `ipldsch_satisfaction.go` is an autogenerated file. That's why I think allowing to provide `-composites=false` to `go vet` would be desirable in this case.

I also tried to see if there is a way to add some magic comment to ignore this file during `go vet` but it seems it does not exist - https://github.com/golang/go/issues/18432#issuecomment-269235329

`go-check-config.json` example:
```
{
  "govet": { 
    "flags": "-composites=false"
  }
}
```

###### Testing

- [x] successfully ran `go generate` with `{"gogenerate":true}` config (`go vet` correctly failed) - https://github.com/galargh/go-ipld-prime-proto/actions/runs/1707665427
- [x] successfully ran `go vet` with `{"govet":{"flags":"-composites=false"}}` config - https://github.com/galargh/go-ipld-prime-proto/actions/runs/1707657893

### DIFF
--- a/templates/.github/workflows/go-check.yml
+++ b/templates/.github/workflows/go-check.yml
@@ -6,7 +6,7 @@ jobs:
     runs-on: ubuntu-latest
     name: All
     env:
-      RUNGOGENERATE: false
+      CONFIG: '{}'
     steps:
       - uses: actions/checkout@v2
         with:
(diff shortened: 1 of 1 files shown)


marten-seemann:
I think @mvdan has been working on fixing these `go vet` warnings. What's the status there?

mvdan:
Yep, the newer go-ipld-prime codegen should generate code with absolutely zero vet warnings. If any are left, they must be fixed. I fixed (hopefully) all of them half a year ago.

The reason go-ipld-prime-proto still has warnings is that it's out of date. It's been deprecated for a while (https://github.com/ipld/go-ipld-prime-proto/issues/23), and at this point we should archive it. It should have already been archived, so my apologies for wasting a little of your time on it :)

I'll go ahead and figure out how we can archive it.

mvdan:
In other words, I don't think we need to support disabling any vet checkers at all.

galargh:
Oh nice! I have missed this issue. That's great news :) 

> It should have already been archived, so my apologies for wasting a little of your time on it :)

No worries at all, it finally gave me an opportunity to look at how `go-check-config.json` is being used in the workflow so I wouldn't count it as a waste of time. 

Closing this PR then.

PR #297, opened 2022-02-23
Run stale repos check on PRs

### LINKED ISSUE #128: Detect missing and misspelled repos
This isn't _critical_, but it would be nice if we could detect missing and misspelled repos in the `config.json`.

galargh:
Resolves #128 

###### Description

This PR enables stale repos check on PRs. The behaviour of running on schedule every 6 hours is not changed.

In PRs, only added repositories are checked.

###### Testing

- [x] https://github.com/protocol/.github/pull/296


.github/workflows/stale-repos.yml
```diff
@@ -1,19 +1,35 @@
...
     name: Find stale repositories
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+      - id: repos
+        name: find repositories to check
+        run: |
+          repos="$(jq '{"\(input_filename)":(.repositories | map(.target))}' configs/*.json | jq -sc 'add')"
+          if [[ '${{ github.event_name }}' == 'pull_request' ]]; then
+            git fetch origin ${{ github.event.pull_request.base.sha }}
```
marten-seemann, line 21: Doesn't the checkout step automatically check out the current state of the repo?

.github/workflows/stale-repos.yml
```diff
@@ -1,19 +1,35 @@
...
+            git reset --hard ${{ github.event.pull_request.base.sha }}
+            base="$(jq '{"\(input_filename)":(.repositories | map(.target))}' configs/*.json | jq -sc 'add')"
```
marten-seemann, line 23: So much jq magic! ;)
A comment explaining what this does would be nice. Same for the next line.

Stebalien: Also, nit: you can do this in one command `jq -nc '[ inputs | ... ] | add`

galargh: Good points - done & done 👍 🙇

.github/workflows/stale-repos.yml
galargh: It checks out a shallow clone (with `--depth=1`). That's why we don't have any info on the PR target commit unless we explicitly fetch it.

marten-seemann: Right. Now I remember fighting this a while back ;)

.github/workflows/stale-repos.yml
```diff
@@ -1,19 +1,47 @@
...
+name: Check Stale Repos
 
 jobs:
   matrix:
     name: Find stale repositories
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+      - id: repos
+        name: find repositories to check
```
marten-seemann, line 17: This step should only run on `pull_request`, right?
We still need to check all repos on `cron`.

.github/workflows/stale-repos.yml
```diff
@@ -1,19 +1,47 @@
...
+            git fetch origin ${{ github.event.pull_request.base.sha }}
+            git reset --hard ${{ github.event.pull_request.base.sha }}
+
+            # retrieves names of target repositories per JSON config file
+            #  on PR target commit
+            base="$(jq -nc '[inputs | {"\(input_filename)":(.repositories | map(.target))}] | add' configs/*.json)"
+
+            # removes names of target repositories that were present in JSON config files
+            #  on PR target commit from the JSON object created on PR head commit
+            repos="$(jq -nc '$head | to_entries | map({"key":"\(.key)","value":(.value - ($base["\(.key)"] // []))}) | from_entries' --argjson head "$repos" --argjson base "$base")"
```
marten-seemann, line 35: That seems a bit convoluted. Isn't there an easier way to diff two jsons?

There are a number of command line tools to do so, and (since we strive to keep 3rd party dependencies to a minimum) there are also numerous ways to use `jq`, e.g. https://stackoverflow.com/questions/31930041/using-jq-or-alternative-command-line-tools-to-compare-json-files.

marten-seemann (changes requested):

.github/workflows/stale-repos.yml
```diff
@@ -1,19 +1,47 @@
...
+        run: |
+          # retrieves names of target repositories per JSON config file
+          # returns a JSON object with the following schema:
+          #  { "$JSON_CONFIG_FILE_PATH": ["$GITHUB_REPOSITORY_FULL_NAME"] }
+          repos="$(jq -nc '[inputs | {"\(input_filename)":(.repositories | map(.target))}] | add' configs/*.json)"
+
+          if [[ '${{ github.event_name }}' == 'pull_request' ]]; then
+            # resets the checkout to PR target commit
+            git fetch origin ${{ github.event.pull_request.base.sha }}
+            git reset --hard ${{ github.event.pull_request.base.sha }}
```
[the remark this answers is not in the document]
galargh: Sure, makes sense!

```suggestion
            git checkout ${{ github.event.pull_request.base.sha }}
```

.github/workflows/stale-repos.yml
galargh: It would have been even clearer if we didn't care about from which config file the repository comes from but I think it's useful info to have in the output.

I gave it another shot here: https://github.com/protocol/.github/pull/297/commits/c67a39d203102811be2a758f4c99adae27ec2bc8

Now this part reads as follows: https://github.com/protocol/.github/blob/c67a39d203102811be2a758f4c99adae27ec2bc8/.github/workflows/stale-repos.yml#L33

.github/workflows/stale-repos.yml
```diff
@@ -1,19 +1,45 @@
...
+
+            # returns a JSON array of repository objects as they exist on pull request target
+            base="$(jq -nc '[inputs | .repositories[] | {"config":input_filename,"repo":.target}]' configs/*.json)"
+
+            # returns a JSON array of repository objects as they exists on pull request head
+            #  without the objects that also exist on pull request target
+            repos="$(jq -nc '$head - $base' --argjson head "$repos" --argjson base "$base")"
+          fi
+
+          echo "::set-output name=this::$repos"
```
marten-seemann, line 36: Nit: `this` might not be the nicest name.

marten-seemann (approved):
Sorry for the very delayed review. LGTM.

Issue #193, opened 2021-09-03
update to Go 1.18


marten-seemann:
The update to Go 1.18 will be an interesting one, as it adds generics. It is safe to assume that many developers will want to use these shiny new language features right away.

This poses a bit of a problem, as using generics requires the Go version in go.mod to be 1.18. However, when we deploy the next Go version bump, the go.mod version will be 1.17 (we always support the most recent two Go versions).

After discussing this with @mvdan offline, we came to the following conclusion:
* People are free to bump the Go version in go.mod manually, if they want to use generics. Having `go-test` test both 1.17 and 1.18 makes sure that we keep our backwards-compatibility guarantees.
* For the Go version update after that (1.19), we'll have to change our upgrade logic in https://github.com/protocol/.github/blob/130b30b68dee0a13da5c74ea398cec3fde640765/.github/workflows/copy-workflow.yml#L109
This can be as simple as grepping for 1.17 in `go-test`.

mvdan:
> For the Go version update after that (1.19), we'll have to change our upgrade logic in

Can you please expand on this, for the sake of clarity? Because the "go.mod version bump" logic on its own will still work just fine. The only bit we'd skip is `go fix ./...`, but luckily that's largely unused these days.

marten-seemann:
Yes, my only concern was that the `if` code block would not be executed for repositories that manually bumped the version in go.mod. What `go fix` will or will not do for Go 1.19 we can't know at this point, can we?

mvdan:
That's true, though the release notes would be pretty clear about whether or not `go fix` starts doing important things again.

marten-seemann:
True, my point is, we don't have to worry about this. Looking at the Go versions in `go-test.yml` will be a more reliable way to detect if we're upgrading a repository, so we can just use that.

mvdan:
That seems fine. My 2c is that it's possible that some users will tinker with those files manually on a temporary basis as needed, like Steven with that msys shell bit, so I don't think it will be as reliable as you think. Maybe we should start leaving behind a sentinel file of some sort, like `.github/.protocol-ci-version` containing a simple version like `go1.17-v1`.

marten-seemann:
Steven’s msys bit would be overwritten the next time the copy workflow runs (e.g. when we add any new repo to config.json).
Our current setup doesn’t really allow _any_ changes to the yml files.

galargh:
I think this one can be closed now since we've updated to Go 1.18 already.

PR #307, opened 2022-03-18
Update to go 1.18

laurentsenta:
Closes https://github.com/protocol/.github/issues/193

Todo:

- [ ] Sync with https://github.com/protocol/.github/pull/261
- [ ] Look into the "change our upgrade logic" discussed in issue 193 and add a note in the yml.


galargh:
I agree with Daniel on `staticcheck`, it'd be less noisy to wait for a week for that new `staticcheck` version.

And with Marten as well, we don't have to wait with merging this PR until `staticcheck` business is sorted out.

marten-seemann (approved):

Issue #258, opened 2021-12-19
Suggestion: drop Go 1.17 when Go 1.18 is released


marten-seemann:
I'd like to suggest to make an exception to our backwards-compatibility guarantee of always supporting the most recent two Go versions. This exception would only cover the transition from Go 1.17 to Go 1.18, when Go 1.19 is released, we'd go back to supporting the two most recent versions (1.18 and 1.19).

### why?

Go 1.18 will add support for generics, which is arguably the biggest (and most anticipated!) language change. I expect that people will want to use these new features as soon as possible. At least I myself can't wait to get my hands dirty!

There are two use cases we need to distinguish here:
1. non-API changes: These will be possible even if we keep the backwards-compatibility guarantee, by using build flags. Ultimately, you'd end up with two copies of the respective code, one that does and one that doesn't use generics.
2. API changes: For some packages, it might be desirable to expose generic types, i.e. interfaces, types and functions that have a type parameter.

I see both of these cases as problematic, to the point of effectively prohibiting the use of generics:
1. Writing generic code, while at the same time keeping a non-generic version around practically duplicates large pieces of code, doubling the maintenance work in case we find a bug in that piece of code (note that the bug might exist in the non-generic or the generic version of the code, or both).
2. Exposing an API that uses generics will be impossible if we still want to support Go 1.17 (unless you're willing to expose different interfaces for Go 1.17 and 1.18, which would make the API **very** hard to use).

### how?

We probably want to start testing using Go 1.18 as soon as the RC is released, which (at the least) means using binaries built using the RC on our own infrastructure. This is not a particularly dangerous thing to do, after all, Google is confident enough in the performance / correctness of Go RCs that they use it for their entire infrastructure. Should that reveal any potential problems with Go 1.18, we could hold off on this plan and decide Go 1.17 for a bit longer, until those problems are fixed.

cc @mvdan @guseggert @aschmahmann @Stebalien @vyzo @BigLep @lidel @raulk

mvdan:
What about making this opt-in? One could opt in by adding `go 1.18` or later to their `go.mod`. Most of our modules right now have `go 1.16`, and can stick with `go 1.17` if they don't want to add generics just yet.

[...]

PR #308, opened 2022-03-21
update to Go 1.18.x

galargh:
Release notice: https://go.dev/blog/go1.18

Open discussions on how to handle Go 1.18: 
- https://github.com/protocol/.github/issues/258
- https://github.com/protocol/.github/issues/193

From what I understand, we haven't reached a consensus on changing the pattern of supporting latest 2 Go versions. That's why I propose this update as-is. 

I'd really appreciate if we could do the next major unified CI release with https://github.com/protocol/.github/pull/261 because it'd give us a chance that the release process itself succeeds without human intervention.

Finally, the `latest` released version of `staticcheck` does not support Go 1.18 yet. The progress is tracked in https://github.com/dominikh/go-tools/issues/1166 and, apparently, there were comments in some other issues that this is going to be complete in about a week. The question is: do we wait for `staticcheck` release or disable it in unified CI for now?

###### Testing
- [x] Updated `.github-test-target` in https://github.com/protocol/.github-test-target/pull/37 (this also showcases the `staticcheck` issue)

### DIFF
--- a/.github/actions/copy-workflow-go/action.yml
+++ b/.github/actions/copy-workflow-go/action.yml
@@ -8,18 +8,18 @@ runs:
       with:
         # This should be the same Go version we use in the go-check workflow.
         # go mod tidy, go vet, staticcheck and gofmt might behave differently depending on the version.
-        go-version: "1.17.x"
+        go-version: "1.18.x"
     - name: bump go.mod go version if needed
       uses: protocol/multiple-go-modules@v1.2
       with:
         working-directory: ${{ env.TARGET_REPO_DIR }}
         run: |
           # We want our modules to support two Go versions at a time.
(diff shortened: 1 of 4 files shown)


mvdan:
I don't think we need to block adding Go 1.18 until the next staticcheck version is out. The main reason that staticcheck is broken with 1.18 is due to generics, but the vast majority of repos using unified CI will keep on supporting Go 1.17, meaning they generally can't use generics yet, unless they go for some funky business with `//go:build go1.18`.

This assumes that it will be painless to update staticcheck in a week's time once they do release proper support for Go 1.18, though. I assume that for most repos it would be a bit bothersome to merge the update to Go 1.18 this week, and a staticcheck update next week. For that reason, it might be best to wait a week until we deploy the Go 1.18 update.

Issue #310, opened 2022-03-30
update staticcheck


marten-seemann:
Staticcheck v0.3.0 was released: https://github.com/dominikh/go-tools/releases/tag/2022.1.

Looks like they also have a dedicated GitHub Action now, so we can simplify our `go-check` a bit. We should probably test that it correctly works with GitHub's code annotations first.

dominikh:
> We should probably test that it correctly works with GitHub's code annotations first.

It should. We emit the necessary magic to install [problem matchers](https://github.com/dominikh/staticcheck-action/blob/master/matchers.json).

galargh:
This was done here: https://github.com/protocol/.github/pull/311

PR #311, opened 2022-03-30
Upgrade staticcheck to v0.3.0

galargh:
It was successfully deployed to https://github.com/protocol/.github-test-target/pull/37

### DIFF
--- a/templates/.github/workflows/go-check.yml
+++ b/templates/.github/workflows/go-check.yml
@@ -24,7 +24,7 @@ jobs:
             echo "RUNGOGENERATE=true" >> $GITHUB_ENV
           fi
       - name: Install staticcheck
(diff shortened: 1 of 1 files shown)


marten-seemann (commented):
What about using the official GitHub Action workflow, see https://github.com/protocol/.github/issues/310?

galargh:
> What about using the official GitHub Action workflow, see #310?

I'm afraid we cannot use https://github.com/dominikh/staticcheck-action together with https://github.com/protocol/multiple-go-modules.

marten-seemann:
Good point. Hadn't thought of that. That's mildly annoying, but not the end of the world.

marten-seemann (approved):

PR #315, opened 2022-04-01
fix: update pull request target to use wildcard

achingbrain:
If `branches` is set to the default branch, it means CI will only run on a PR that targets the default branch.

Sometimes we have release branches like `/release/v1.0.0` which use use to backport fixes to previous versions - we still want CI to run on these backport PRs so change the PR branch target to run CI on PRs that target any branch.


galargh (approved):
I let myself delete the branch filter altogether since we want it to match all anyway (https://github.com/protocol/.github/pull/315/commits/cfb2907c99f2e7304cb3a7f7827655f748183115). Hope you don't mind ;)

PR #318, opened 2022-04-04
fix issues uncovered during 2022-04-04 release

galargh:
The changes here make it possible to proceed with the 2022-04-04 release. Merging immediately to unblock.

### DIFF
--- a/.github/actions/copy-workflow-go/action.yml
+++ b/.github/actions/copy-workflow-go/action.yml
@@ -31,6 +31,9 @@ runs:
             # Update the version in go.mod. This alone ensures there's a diff.
             go mod edit -go $TARGET_VERSION
 
+            # We tidy immediately because some updates require it
+            go mod tidy
+
             # In the future, "go fix" may make changes to Go code,
             # such as to adapt to language changes or API deprecations.
             # This is largely a no-op as of Go 1.17, and that's fine.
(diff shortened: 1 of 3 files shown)


.github/actions/copy-workflow-go/action.yml
```diff
@@ -31,6 +31,9 @@ runs:
             # Update the version in go.mod. This alone ensures there's a diff.
             go mod edit -go $TARGET_VERSION
 
+            # We tidy immediately because some updates require it
+            go mod tidy
+
```
galargh, line 36: See https://github.com/protocol/.github/runs/5815591028?check_suite_focus=true for example

.github/workflows/create-prs.yml
```diff
@@ -0,0 +1,73 @@
+name: Create PRs
```
galargh, line 1: A script that should help me recover when branch pushes succeeded but the PR creation step failed. To be debugged after the merge.

.github/workflows/dispatch.yml
```diff
@@ -136,8 +138,7 @@ jobs:
               echo "Failed to create a PR for '$target'"
               failed+=("$target(create)")
             fi
-            # 3 seconds seemed not to be enough, trying 10
-            sleep 10
+            sleep 3
```
galargh, line 141: It doesn't seem to make a difference here so let's stick with 3.

PR #323, opened 2022-04-06
Extract PR creation out of Dispatch workflow

galargh:
In this PR I remove PR creation phase out of Dispatch workflow entirely. Instead, I add a new trigger to the Create PRs workflow - `workflow_run`. With this Create PRs is triggered whenever Dispatch workflow finishes.

I also fix Create PRs workflow. Here's how it works after the changes:
1. Checkout `.github` at the same ref Dispatch workflow was using or at the branch that was chosen when dispatching the run manually.
1. Collect all the repositories from JSON configs. 
1. For each repository:
    1. Check if `web3-bot/sync` branch exists and if it is `ahead` of the default branch. If not, skip processing this repository.
    1. Check if an open PR from `web3-bot/sync` to the default branch exists. If not, skip processing this repository.
    1. Try:
        1. Create PR from `web3-bot/sync` to the default branch. 
        1. If successful, sleep for 1 second and stop trying.
        1. If failed with secondary rate limit error message and tried less than 12 times, sleep for INTERVAL (starts at 1) seconds, double the interval and try again.
        1. If failed with other reason or exceeded try attempt limit, warn about failing to create a PR for this repository.
        
The `Try` loop in the Create PRs workflow can take up to 70 minutes. It is not documented anywhere but from experience, I have seen us being locked out of PR creation for 30 minutes.

###### Testing

I have used Create PRs workflow during 2022-04-04 release.

### DIFF
--- a/.github/workflows/create-prs.yml
+++ b/.github/workflows/create-prs.yml
@@ -24,15 +24,7 @@ jobs:
         run: |
           targets=()
           for config in configs/*.json; do
-            if [[ $GITHUB_REF == refs/heads/testing && $config != configs/testing.json ]]; then
-              continue
-            fi
-            if [[ $GITHUB_REF != refs/heads/testing && $config == configs/testing.json ]]; then
-              continue
-            fi
-            echo "::group::$config"
             targets+=($(jq -rc ".repositories[] | .target" $config))
-            echo "::endgroup::"
           done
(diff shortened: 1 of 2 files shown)


laurentsenta (approved):
👍  thanks for the thorough description,

PR #327, opened 2022-04-10
run go mod tidy with -compat=1.17 option

### LINKED ISSUE #326: go-check failing
```
❯ go mod tidy
go: downloading github.com/libp2p/go-libp2p-core v0.14.1-0.20220410093848-f927693924b0
github.com/libp2p/go-libp2p-yamux imports
        github.com/libp2p/go-libp2p-core/network tested by
        github.com/libp2p/go-libp2p-core/network.test imports
        github.com/stretchr/testify/require imports
        github.com/stretchr/testify/assert imports
        gopkg.in/yaml.v3 tested by
        gopkg.in/yaml.v3.test imports
        gopkg.in/check.v1 loaded from gopkg.in/check.v1@v0.0.0-20161208181325-20d25e280405,
        but go 1.16 would select v1.0.0-20180628173108-788fd7840127

To upgrade to the versions selected by go 1.16:
        go mod tidy -go=1.16 && go mod tidy -go=1.17
If reproducibility with go 1.16 is not needed:
        go mod tidy -compat=1.17
For other options, see:
        https://golang.org/doc/modules/pruning
```

Not sure I fully understand the root of the error here, but maybe we should use `go mod tidy -compat=1.17` instead of `go mod tidy` in `go-check`.

marten-seemann:
Fixes #326.

Note that this is targeting `master`. `next` seems to be outdated (?).

### DIFF
--- a/templates/.github/workflows/go-check.yml
+++ b/templates/.github/workflows/go-check.yml
@@ -29,7 +29,7 @@ jobs:
         uses: protocol/multiple-go-modules@v1.2
         with:
           run: |
-            go mod tidy
+            go mod tidy -compat=1.17
             if [[ -n $(git ls-files --other --exclude-standard --directory -- go.sum) ]]; then
               echo "go.sum was added by go mod tidy"
               exit 1


galargh (changes requested):
Let's try to understand the issue first. In the meantime, how about we use the workaround suggested by `go mod tidy` itself i.e. `go mod tidy -go=1.16 && go mod tidy -go=1.17 `, e.g. https://github.com/libp2p/go-libp2p-yamux/pull/52#issuecomment-1094669625?

I remember I did have to do it for one more module already, I'll edit it with its' name when I find it so that we have more data points to analyse.

marten-seemann:
Ok, so running `go mod tidy -go=1.16 && go mod tidy -go=1.17` works, in the sense that after that a simple `go mod tidy` doesn't introduce any changes after that.

Only the other way round (`go mod tidy -compat=1.17` and then `go mod tidy`) reproduces the error. And apparently that's the only thing I had tried yesterday :(

I guess this resoles the issue. Sorry for the noise.

galargh:
For completeness, I think that the difference between Go 1.16 and Go 1.17 dependency resolution in this case comes from [lazy loading](https://go.dev/ref/mod#lazy-loading).

PR #332, opened 2022-05-02
allow optional ADDITIONAL_TEST_OPTIONS env var to ammend `go test` command

rvagg:
Ref: https://github.com/ipld/go-ipld-prime/pull/409

or ... got a better idea or a more elegant way to achieve this kind of thing?

1. I can't use `-coverpkg` and run vanilla tests (https://github.com/golang/go/issues/23410, https://github.com/ipld/go-ipld-prime/blob/master/schema/gen/go/HACKME_testing.md)
2. I can turn off plugins to make it work but it skips some important tests in the process and it needs a build tag to be passed in `go test`, hence this change
3. I can add those tests back in with an additional custom workflow
4. I don't really care that much about coverage at the moment, highest priority is to get proper CI running, another way forward here might be an option to just turn off coverage entirely

### DIFF
--- a/templates/.github/workflows/go-test.yml
+++ b/templates/.github/workflows/go-test.yml
@@ -40,7 +40,9 @@ jobs:
           # Use -coverpkg=./..., so that we include cross-package coverage.
           # If package ./A imports ./B, and ./A's tests also cover ./B,
           # this means ./B's coverage will be significantly higher than 0%.
-          run: go test -v -coverprofile=module-coverage.txt -coverpkg=./... ./...
+          run: |
+            additional_opts=( $ADDITIONAL_TEST_OPTIONS )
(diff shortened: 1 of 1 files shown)


galargh:
If we just wanna pass additional tag to `go test`, I think we could use `GOFLAGS` for this. Here's an example where we extend timeout for tests -> https://github.com/libp2p/go-libp2p-pubsub-router/blob/292d99457d224853706c5e49f8ddc112740a856a/.github/actions/go-test-setup/action.yml

galargh:
It'd look something like this: https://github.com/ipld/go-ipld-prime/pull/410

rvagg:
solved with https://github.com/ipld/go-ipld-prime/pull/410, thanks

PR #333, opened 2022-05-03
Nits: Improve if statements

ajnavarro:
use `always()` instead of `(success() || failure())`

Signed-off-by: Antonio Navarro Perez <antnavper@gmail.com>

### DIFF
--- a/templates/.github/workflows/go-check.yml
+++ b/templates/.github/workflows/go-check.yml
@@ -36,7 +36,7 @@ jobs:
             fi
             git diff --exit-code -- go.sum go.mod
       - name: gofmt
-        if: ${{ success() || failure() }} # run this step even if the previous one failed
+        if: always() # run this step even if the previous one failed
         run: |
           out=$(gofmt -s -l .)
           if [[ -n "$out" ]]; then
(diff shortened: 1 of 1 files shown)


templates/.github/workflows/go-check.yml
```diff
@@ -36,28 +36,28 @@ jobs:
...
+        if: always() # run this step even if the previous one failed
         uses: protocol/multiple-go-modules@v1.2
         with:
           run: |
             set -o pipefail
             staticcheck ./... | sed -e 's@\(.*\)\.go@./\1.go@g'
       - name: go generate
         uses: protocol/multiple-go-modules@v1.2
-        if: (success() || failure()) && env.RUNGOGENERATE == 'true'
+        if: env.RUNGOGENERATE == 'true'
```
marten-seemann, line 60: This is wrong. This should also run if the previous step failed.

ajnavarro: Woops, true. changed.

galargh:
FYI, `always()` is equivalent to `success() || failure() || cancelled()` so this is in fact changing the previous behaviour.

ajnavarro:
Feel free to close this PR if the new behavior is not correct.

galargh:
> Feel free to close this PR if the new behavior is not correct.

Unless this is specifically requested, I think it'd be valuable to preserve the ability to cancel a workflow run and have it quit immediately. Closing for now then.

PR #340, opened 2022-05-25
Upgrade codecov-action to v3.1.0

galargh:
Following up on https://discord.com/channels/806902334369824788/979025498816520292/979046186143072296


marten-seemann:
Trying it out in https://github.com/libp2p/go-libp2p/pull/1558.

marten-seemann:
Looks like this doesn't help, see https://github.com/codecov/codecov-action/issues/745.

galargh:
There was another update (I assume on the codecov server side) after the last testing session - https://github.com/codecov/codecov-action/issues/745#issuecomment-1139083047

There is no newer version we can't upgrade to so I think we should go ahead with this one and hope for the best 🤞

laurentsenta (approved):

PR #341, opened 2022-05-26
fix: update code-cov action

achingbrain:
Updates code-cov action to commit tagged as [3.1.0](https://github.com/codecov/codecov-action/releases/tag/v3.1.0)

### DIFF
--- a/templates/.github/workflows/js-test-and-release.yml
+++ b/templates/.github/workflows/js-test-and-release.yml
@@ -35,7 +35,7 @@ jobs:
           node-version: ${{ matrix.node }}
       - uses: ipfs/aegir/actions/cache-node-modules@master
       - run: npm run --if-present test:node
-      - uses: codecov/codecov-action@f32b3a3741e1053eb607407145bc9619351dc93b # v2.1.0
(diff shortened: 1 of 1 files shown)


marten-seemann:
Is this intended to fix the OSX upload errors? I don't think this will work: https://github.com/protocol/.github/pull/340#issuecomment-1138429610

achingbrain:
No, but likely the first thing they will say is "Are you running the latest version?" 😆

marten-seemann:
I tried out the latest version in https://github.com/libp2p/go-libp2p/pull/1558, and it still doesn't work.
I'll wait for a reply of the Codecov team before taking any action on the Go side.

achingbrain:
https://github.com/codecov/codecov-action/issues/745 has been closed and it also seems to be fixed.

Any further objections to this PR?

galargh (approved):

achingbrain:
Reading their docs `fail_ci_if_error` defaults to `false` so setting it explicitly should be unnecessary, though, you know, it didn't stop https://github.com/codecov/codecov-action/issues/745 so maybe that setting does nothing 🤷

PR #355, opened 2022-06-30
feat: add a weekly scheduled action which adds Unified CI to new Go repos

### LINKED ISSUE #23: Auto-deploy CI to dependencies when they're added
We'll occasionally make some fancy new library and add it as a dependency to one of our projects but fail to actually setup CI.

We should add a action that:

1. Checks that all (transitive) go.mod dependencies housed in one of our orgs has CI correctly configured.
2. If not, fails CI, and makes a PR to _deploy_ CI to that repo.

This should help prevent this from happening in the future.

galargh:
Resolves #23

This PR adds a weekly scheduled workflow which:
1. Finds all Go repositories without Unified CI from orgs that web3-bot has access to
1. Downloads a list of Go repositories without Unified CI that was created by a previous run of the workflow
1. Determines which repositories are new
1. Adds the new repositories to `configs/go.json`
1. Creates a PR and tags most active contributors of the repositories being added

I decided to implement the above as a [TypeScript Action](https://github.com/actions/typescript-action). It comes with a bit of boilerplate but I'd say the end result is much more approachable and better tested.


galargh:
Let's think either about automating `npm run all` on PRs or changing it into a docker action which knows how to run ts.

laurentsenta (approved):
lgtm, ts-node discussion, can be dealt with later

Issue #362, opened 2022-08-03
update to Go 1.19


marten-seemann:
Go 1.19 was released yesterday: https://go.dev/blog/go1.19 🎉

This means we can update CI to use Go 1.18 and 1.19, and drop 1.17, which means we can (finally!) make use of generics!

As far as I can see, these are the steps we need to take:
- [x] update Go version in workflow files
- [x] update staticcheck (open question: will there be a Go 1.19 update?)
- [x] update Codecov Action (because as we all know, uploading a code coverage file is one of the hard problems in computer science)
- [x] include https://github.com/protocol/.github/pull/353, which uses the `-shuffle` flag on `go test` (it's already merged into `next`)

Did I forget anything?

aschmahmann:
Note that generally speaking we wait for 1.X.1 given 1.X tends to have a pile of bugs. That being said it's probably fine to update CI at 1.X as long as the binaries that want it be more careful (e.g. kubo or anything built via https://github.com/ipfs/distributions) keep deploying with 1.X-1.Latest.

marten-seemann:
@aschmahmann Since Unified CI is only dealing with tests (and not deployments / builds), this shouldn’t have any effect on kubo at all, right? And since we’re always testing the latest two Go versions, the Go version you’re using for releasing would still be tested (even assuming that kubo was using Unified CI, which I thought it isn’t yet?).

galargh:
Linking https://github.com/protocol/.github/issues/317 as a template for how to perform a release.

#### TODOs breakdown

###### Codecov

The newest release of codecov action is still `3.1.0`. 

Let's go ahead with https://github.com/protocol/.github/pull/340 for this then. 

Their second comment here https://github.com/codecov/codecov-action/issues/745#issuecomment-1139083047 happened after testing in https://github.com/libp2p/go-libp2p/pull/1558 so there's a chance that `3.1.0` will resolve the issue for us.

###### staticcheck

I haven't seen any chatter on this beyond https://github.com/dominikh/go-tools/issues/1306

Maybe the latest version already works with Go 1.19? We'll need to check this out.

###### update Go version in workflow files

I'll try to prepare this tomorrow.

#### Timeline

I think I should be able to prepare the following by the end of this week:
- merge https://github.com/protocol/.github/pull/340
- prepare PR updating Go version in uCI workflows
- prepare PR upgrading staticcheck
- try out statichceck with Go 1.19 in a couple of repos
- prepare release announcement

[...]

PR #365, opened 2022-08-04
feat: enable support for opting in to technical preview updates

galargh:
###### Description

This PR adds support for a new property `source_ref` in the repository targets. When `source_ref` is set, `Deploy` workflow will check out `protocol/.github` at said ref and copy the workflows from there instead of `github.ref` (the branch in which the workflow is running). 

If `source_ref` is not specified, `Deploy` workflow will act exactly as it did before i.e. it will check out `github.ref`.

###### Why now?

We'd like to let users opt-in to Go 1.19 upgrade before we're able to perform the official release (https://github.com/protocol/.github/issues/362).

But I think this will become useful beyond this immediate need, e.g. for continuous testing of the `next` branch. 

###### Testing
- [x] push to `testing` without setting `source_ref` on `protocol/.github-test-target ` - https://github.com/protocol/.github/runs/7674190338?check_suite_focus=true
```
Run actions/checkout@v2
  with:
    path: template-repo
    ref: refs/heads/testing
```
- [x] push to `testing` with `source_ref` set to `next` on `protocol/.github-test-target ` - https://github.com/protocol/.github/runs/7674236444?check_suite_focus=true
```
Run actions/checkout@v2
  with:
    path: template-repo
    ref: next
```

### DIFF
--- a/.github/workflows/copy-workflow.yml
+++ b/.github/workflows/copy-workflow.yml
@@ -39,6 +39,7 @@ jobs:
       uses: actions/checkout@v2
       with:
         path: ${{ env.TEMPLATE_REPO_DIR }}
+        ref: ${{ matrix.cfg.source_ref || github.ref }}
     - name: determine GitHub default branch
       working-directory: ${{ env.TARGET_REPO_DIR }}
       run: |
--- a/.github/workflows/dispatch.yml
+++ b/.github/workflows/dispatch.yml
@@ -7,6 +7,7 @@ name: Dispatch
 on:
   push:
     branches: [ master, testing ]
+  workflow_dispatch:
 
 env:
   # Number of repositories in a batch.
@@ -23,6 +24,7 @@ env:
 
 jobs:
   matrix:
+    if: github.ref == 'refs/heads/master' || github.ref == 'refs/heads/testing'
     name: Batch targets
     runs-on: ubuntu-latest
     outputs:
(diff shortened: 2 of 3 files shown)


galargh:
I just thought of one more thing that we should have for this - triggering. Right now, dispatch doesn't trigger on pushes to `next`. And we cannot simply add `next` to https://github.com/protocol/.github/blob/d36759d161d15a9e3c4e018d3a79227e469b8354/.github/workflows/dispatch.yml#L9 because then `dispatch` would kick off `deploy`s with `refs/heads/next`` for `github.ref`.

Maybe, at least for now, it'd be enough to add `workflow_dispatch` trigger to the `dispatch` workflow. We could guard it so that `dispatch` quits immediately if `github.ref` is different than `refs/heads/master` or `refs/heads/testing`.

With that setup, when someone creates a PR to opt-in their repos to updates from a different branch, then it gets the first update immediately after the merge to `master`. The subsequent upgrades happen either when `master` moves or Unified CI maintainers decide to kick off the `dispatch` workflow.

galargh:
I added the `workflow_dispatch` trigger in https://github.com/protocol/.github/pull/365/commits/5e0b0b9be23401ac6b5ac72bfe8513b8998105fc and here's a run on the `testing` branch that was let through the conditional checking `github.ref` https://github.com/protocol/.github/runs/7675207528?check_suite_focus=true

laurentsenta (approved):

Issue #367, opened 2022-08-07
Do "Technical Previews" make sense?


marten-seemann:
I see a new feature added called "Technical Previews", and I'm questioning how useful it is. If a repo opts in, it will receive all updates merged into the `next` branch, instead of `master`. I wouldn't want to enable it for any repo that I'm maintaining, as the normal web3-bot updates are already pretty noisy.

I do see the need to test updates before doing a deployment run to all repos. Maybe the following would be more useful: Repos could opt-in to _testing_ updates (without merging to master!) before we run the deployment.
Essentially, the suggestion is to have 3 branches:
1. `next`: nothing happens when a PR is merged here
1. `deploy` (feel free to suggest a better name): once this branch is moved forward, repos that opted in will receive a PR from web3-bot, but without the _automerge_ job. This would allow us to test the new version of the workflows before triggering the org-wide deployment, while also creating minimal noise.
1. `master`: this works as today. When this branch is changed, this triggers a deployment to all repos, and PRs will be automerged.

@laurentsenta @galargh wdyt?

laurentsenta:
Having a dedicated pre-production branch with a release frequency similar to master's makes sense.

A few notes:

- We could try this now if you want: the `source_ref` config can be any branch.
- When it comes to naming, something like `develop > next > master` seems more explicit. It's easier to tell the nature and frequency of updates my project will receive.
- If technical previews are NOT auto merged, wouldn't that create more maintenance work for you? (you'd have a new PR to babysit, keep track of vulnerability fixes, etc.).

marten-seemann:
> When it comes to naming, something like develop > next > master seems more explicit. It's easier to tell the nature and frequency of updates my project will receive.

Nice suggestion, that's a lot better!

> * If technical previews are NOT auto merged, wouldn't that create more maintenance work for you? (you'd have a new PR to babysit, keep track of vulnerability fixes, etc.).

In my mind, I don't want to have to deal with CI **at all**. I just want to be sure that it runs with the up-to-date configuration. That's why I started the Unified CI effort back then. That's why I wouldn't want to opt into Technical Previews for any of the repos (that I maintain).

I do however see that a new Unified CI configuration should be tested before a deployment to hundreds of repos is triggered. Ideally, this would be invisible to the repo maintainers. Maybe a branch would be better suited than the draft PR I suggested above. There'd probably be some tooling needed to check that the builds on that branch actually succeed. Maybe a check-list for manual checks would be good enough for the start though.

[...]

PR #369, opened 2022-08-12
feat: add go-libp2p to preview

laurentsenta:
What I did:

1. create the branch `preview` from `next`: https://github.com/protocol/.github/tree/preview ([diff](https://github.com/protocol/.github/compare/next...preview))
2. update `source_ref` in `go-libp2p`

Merging into master will trigger a deployment of the `preview` version.

Taking into account the discussion in https://github.com/protocol/.github/issues/367 we use a branch that is not `next` until we agree on a definitive naming.

### DIFF
--- a/configs/go.json
+++ b/configs/go.json
@@ -343,7 +343,8 @@
       "target": "libp2p/go-flow-metrics"
     },
     {
-      "target": "libp2p/go-libp2p"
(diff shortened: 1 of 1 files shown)


marten-seemann (approved):
This looks pretty straightforward!

PR #373, opened 2022-08-23
chore: removed archived repos from go config

galargh:
Fixing issues raised by https://github.com/protocol/.github/runs/7972743573?check_suite_focus=true


marten-seemann (approved):
I can confirm that these are the 4 repos that we recently moved into go-libp2p.

laurentsenta (approved):

PR #376, opened 2022-08-24
chore: update setup-node to v3

galargh:
AFAICT the only breaking change is removal of `version` prop in favor of `node-version` but we're using the latter already - https://github.com/actions/setup-node/releases.


laurentsenta (approved):
👍

PR #366, opened 2022-08-05
Unified CI Release - 2022-08-23

galargh:

PR #377, opened 2022-08-25
automate cleaning up deprecated io/ioutil

galargh:
Following up on https://github.com/protocol/.github/pull/366#issuecomment-1225751649

I tested the newly added part locally on `ipfs/go-ipfs-files` (note, that particular repo will require manual intervention anyway because it uses `ioutil.ReadDir` too but I think most other repos don't). 

Merging to `master` because it will only affect the repos with unmerged upgrades to Go 1.19.

### DIFF
--- a/.github/actions/copy-workflow-go/action.yml
+++ b/.github/actions/copy-workflow-go/action.yml
@@ -49,24 +49,23 @@ runs:
 
             # As of Go 1.19 io/ioutil is deprecated
             # We automate its upgrade here because it is quite a widely used package
-            if [[ $TARGET_VERSION == "1.18" ]]; then
-              while read file; do
-                sed -i 's/ioutil.NopCloser/io.NopCloser/' "${file}";
-                sed -i 's/ioutil.ReadAll/io.ReadAll/' "${file}";
-                # Skipping ReadDir replacement because it's a bit more complicated
-                # See https://pkg.go.dev/io/ioutil#ReadDir
-                # sed -i 's/ioutil.ReadDir/os.ReadDir/' "${file}";
-                sed -i 's/ioutil.ReadFile/os.ReadFile/' "${file}";
-                sed -i 's/ioutil.TempDir/os.MkdirTemp/' "${file}";
-                sed -i 's/ioutil.TempFile/os.CreateTemp/' "${file}";
-                sed -i 's/ioutil.WriteFile/os.WriteFile/' "${file}";
-              done <<< "$(find . -type f -name '*.go')"
+            while read file; do
+              sed -i 's/ioutil.NopCloser/io.NopCloser/' "${file}";
+              sed -i 's/ioutil.ReadAll/io.ReadAll/' "${file}";
+              # Skipping ReadDir replacement because it's a bit more complicated
+              # See https://pkg.go.dev/io/ioutil#ReadDir
+              # sed -i 's/ioutil.ReadDir/os.ReadDir/' "${file}";
+              sed -i 's/ioutil.ReadFile/os.ReadFile/' "${file}";
+              sed -i 's/ioutil.TempDir/os.MkdirTemp/' "${file}";
+              sed -i 's/ioutil.TempFile/os.CreateTemp/' "${file}";
+              sed -i 's/ioutil.WriteFile/os.WriteFile/' "${file}";
+            done <<< "$(find . -type f -name '*.go')"
 
-              goimports -w .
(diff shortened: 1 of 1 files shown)


.github/actions/copy-workflow-go/action.yml
```diff
@@ -46,6 +46,27 @@ runs:
             # We don't tidy, because the next step does that.
             # Separate commits also help with reviews.
             git commit -m "bump go.mod to Go $TARGET_VERSION and run go fix"
+
+            # As of Go 1.19 io/ioutil is deprecated
+            # We automate its upgrade here because it is quite a widely used package
+            if [[ $TARGET_VERSION == "1.18" ]]; then
```
marten-seemann, line 52: This `if` is always true, isn't it?

galargh: That's correct. I wanted to make it clear that we can remove it in time for the next upgrade but I guess a comment might be more fitting for that.

marten-seemann: We probably shouldn't remove it. If any of the updates is not merged (and the repo stays on an old uCI version), it would be nice if this was included in the next update, just in case. wdyt?

galargh: You're right, we might also want to enrol a new repository that's not on latest Go yet and still uses `ioutil` - it'd come in handy then too.

Let me just remove the if-clause then.

.github/actions/copy-workflow-go/action.yml
```diff
@@ -49,24 +49,23 @@ runs:
...
+              # See https://pkg.go.dev/io/ioutil#ReadDir
+              # sed -i 's/ioutil.ReadDir/os.ReadDir/' "${file}";
+              sed -i 's/ioutil.ReadFile/os.ReadFile/' "${file}";
+              sed -i 's/ioutil.TempDir/os.MkdirTemp/' "${file}";
+              sed -i 's/ioutil.TempFile/os.CreateTemp/' "${file}";
+              sed -i 's/ioutil.WriteFile/os.WriteFile/' "${file}";
+            done <<< "$(find . -type f -name '*.go')"
 
-              goimports -w .
+            go install golang.org/x/tools/cmd/goimports@b3b5c13b291f9653da6f31b95db100a2e26bd186 # v0.1.12
```
marten-seemann, line 64: When using `go install`, it's safe to use `@version` directly. The Go module proxy makes sure the hash doesn't change.

.github/actions/copy-workflow-go/action.yml
```diff
@@ -46,6 +46,26 @@ runs:
...
+              # Skipping ReadDir replacement because it's a bit more complicated
+              # See https://pkg.go.dev/io/ioutil#ReadDir
+              # sed -i 's/ioutil.ReadDir/os.ReadDir/' "${file}";
+              sed -i 's/ioutil.ReadFile/os.ReadFile/' "${file}";
+              sed -i 's/ioutil.TempDir/os.MkdirTemp/' "${file}";
+              sed -i 's/ioutil.TempFile/os.CreateTemp/' "${file}";
+              sed -i 's/ioutil.WriteFile/os.WriteFile/' "${file}";
+            done <<< "$(find . -type f -name '*.go')"
+
+            go install golang.org/x/tools/cmd/goimports@b3b5c13b291f9653da6f31b95db100a2e26bd186 # v0.1.12
```
galargh, line 64: Got the id from https://cs.opensource.google/go/x/tools/+/b3b5c13b291f9653da6f31b95db100a2e26bd186...refs/tags/v0.1.12

.github/actions/copy-workflow-go/action.yml
galargh: Cool! Let's go with this then:
```suggestion
            go install golang.org/x/tools/cmd/goimports@v0.1.12
```

marten-seemann (approved):
LGTM. Thank you @galargh!

Does it make sense to try this out with one or two repos first?

galargh:
> Does it make sense to try this out with one or two repos first?

Definitely! I was going to use `testing` branch for that - will merge if everything looks good.

galargh:
Tested on:
- https://github.com/multiformats/go-multiaddr/pull/180 (successfully merged now)
- https://github.com/ipfs/go-ipfs-files/pull/55 (still failing because it uses `ReadDir` but otherwise looks OK)

PR #378, opened 2022-08-25
hide goimport fix commit behind git diff check

galargh:
`git commit` exists with `1` if it has nothing to commit - that's why we need to check if there's anything to commit ahead of time.


.github/actions/copy-workflow-go/action.yml
```diff
@@ -59,13 +59,16 @@ runs:
               sed -i 's/ioutil.TempDir/os.MkdirTemp/' "${file}";
               sed -i 's/ioutil.TempFile/os.CreateTemp/' "${file}";
               sed -i 's/ioutil.WriteFile/os.WriteFile/' "${file}";
+              sed -i 's/ioutil.Discard/io.Discard/' "${file}";
```
galargh, line 62: Missed this replacement previously - adding it here to avoid 2 deploys.

PR #382, opened 2022-08-30
Add js-multiformats to unified-ci

RangerMauve:
This is part of the JS IPLD ESM migration: https://github.com/ipld/ipld/issues/224

Do we need to merge the [aegir changes](https://github.com/multiformats/js-multiformats/pull/196) into js-multiformtats before or after this PR gets merged?

cc @achingbrain @rvagg


rvagg:
I think we can do it either way. My preference has been to get the repo set up with a current(ish) version of what unified-ci sets up and then register it here and get it to sync just to avoid the noise.

BigLep:
This won't get merged until after https://github.com/multiformats/js-multiformats/pull/199 is merged.

PR #383, opened 2022-09-01
feat: replace ioutil.ReadDir too

galargh:
I looked through a couple of PRs that are still having issues with `ioutil` and it turns out that this simple replacement would work there because we only care about file names and the `os.ReadDir` results do offer that.

### DIFF
--- a/.github/actions/copy-workflow-go/action.yml
+++ b/.github/actions/copy-workflow-go/action.yml
@@ -52,9 +52,8 @@ runs:
             while read file; do
               sed -i 's/ioutil.NopCloser/io.NopCloser/' "${file}";
               sed -i 's/ioutil.ReadAll/io.ReadAll/' "${file}";
-              # Skipping ReadDir replacement because it's a bit more complicated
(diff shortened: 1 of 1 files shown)


marten-seemann (commented):
I think this needs to be fixed manually, since the return value is different:
> As of Go 1.16, os.ReadDir is a more efficient and correct choice: it returns a list of fs.DirEntry instead of fs.FileInfo, and it returns partial results in the case of an error midway through reading a directory.

galargh:
Yes, I know. I was just saying that in the cases I looked at so far we were only interested `fs.FileInfo.Name()` and `fs.DirEntry.Name()` exists too.

`IsDir()` is also common between the two.

So there are some cases in which the replacement will just work. It will fail if we're trying to access any of `Size()`, `Mode()`, `ModTime()`, `Sys()`. They don't exist on `DirEntry` so it should be caught early too.

marten-seemann (approved):
Makes sense, thanks for the explanation!

PR #395, opened 2022-09-21
Add js-mafmt to unified ci

achingbrain:


galargh (approved):
Looks good :) 

I double-checked if `web3-bot` has push access to the repo: https://github.com/multiformats/github-mgmt/blob/master/github/multiformats.yml#L745

PR #412, opened 2022-11-08
Add option to skip `32-bit` go test

### LINKED ISSUE #388: Should we support opting out of 32bit tests in go-test workflow?
@masih raised this in https://filecoinproject.slack.com/archives/C03KLC57LKB/p1662663509578329

masih:
Introduce an option to configure `go-test` to allow completely skipping `32-bit` tests.

Fixes #388

### DIFF
--- a/templates/.github/workflows/go-test.yml
+++ b/templates/.github/workflows/go-test.yml
@@ -10,6 +10,7 @@ jobs:
         go: [ "1.18.x", "1.19.x" ]
     env:
       COVERAGES: ""
+      SKIP32BIT: false
     runs-on: ${{ format('{0}-latest', matrix.os) }}
     name: ${{ matrix.os }} (go ${{ matrix.go }})
     steps:
@@ -34,6 +35,13 @@ jobs:
       - name: Run repo-specific setup
         uses: ./.github/actions/go-test-setup
         if: hashFiles('./.github/actions/go-test-setup') != ''
+      - name: Read config
+        if: hashFiles('./.github/workflows/go-test-config.json') != ''
+        shell: bash
+        run: |
+          if jq -re .skip32bit ./.github/workflows/go-test-config.json; then
+            echo "SKIP32BIT=true" >> $GITHUB_ENV
+          fi
       - name: Run tests
         uses: protocol/multiple-go-modules@v1.2
         with:
@@ -42,7 +50,7 @@ jobs:
           # this means ./B's coverage will be significantly higher than 0%.
           run: go test -v -shuffle=on -coverprofile=module-coverage.txt -coverpkg=./... ./...
       - name: Run tests (32 bit)
-        if: ${{ matrix.os != 'macos' }} # can't run 32 bit tests on OSX.
+        if: matrix.os != 'macos' && env.SKIP32BIT == 'false' # can't run 32 bit tests on OSX.
         uses: protocol/multiple-go-modules@v1.2
         env:
           GOARCH: 386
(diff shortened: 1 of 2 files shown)


templates/.github/workflows/go-test.yml
```diff
@@ -42,7 +49,7 @@ jobs:
           # this means ./B's coverage will be significantly higher than 0%.
           run: go test -v -shuffle=on -coverprofile=module-coverage.txt -coverpkg=./... ./...
       - name: Run tests (32 bit)
-        if: ${{ matrix.os != 'macos' }} # can't run 32 bit tests on OSX.
+        if: ${{ matrix.os != 'macos' }} && env.SKIP32BIT == 'false' # can't run 32 bit tests on OSX.
```
marten-seemann, line 52: Why is this not in the `${{ ... }}`? Is the env a string or a boolean?

templates/.github/workflows/go-test.yml
```diff
@@ -10,6 +10,7 @@ jobs:
         go: [ "1.18.x", "1.19.x" ]
     env:
       COVERAGES: ""
+      SKIP32BIT: false
```
marten-seemann, line 13: Is this a string or a boolean?

marten-seemann (commented):
This should probably target the `next` branch, not `master`.

masih:
@marten-seemann based on a quick search, I can't find documentation on weather environment variables with boolean-looking value are interpreted as `string` or `bool` by GitHub Actions template. I would expect environment variable to be interpreted as `string`?

Considering the approach in this PR is consistent with the way by which other boolean config is parsed elsewhere (e.g. `gogenerate` config), would it make sense to move revising the boolean vs string interpretation outside of the scope of this PR?

marten-seemann:
I'm not really concerned about consistency here. @galargh might be though ;)
I'd recommend trying out this exact workflow in a branch in your repo, to make sure it actually works. I remember that the string vs. boolean mess has caused me some headache in the past.

galargh:
> @marten-seemann based on a quick search, I can't find documentation on weather environment variables with boolean-looking value are interpreted as `string` or `bool` by GitHub Actions template. I would expect environment variable to be interpreted as `string`?

That's correct. If it's in `env`, it's a string.

> I'd recommend trying out this exact workflow in a branch in your repo, to make sure it actually works. I remember that the string vs. boolean mess has caused me some headache in the past.

Yes, please! Testing it out on a real repo would be extremely helpful. 

> This should probably target the `next` branch, not `master`.

Let's make this change too. We don't want to release it immediately to all the repos. If you want it to reach your repo earlier than the next major release, you can use technical preview feature - https://github.com/protocol/.github#technical-preview

templates/.github/workflows/go-test.yml
galargh: That would be a string.

templates/.github/workflows/go-test.yml
galargh: The check is correct. We might need to adjust the templating though.

Either with:
```suggestion
        if: matrix.os != 'macos' && env.SKIP32BIT == 'false' # can't run 32 bit tests on OSX.
```
or: 
```suggestion
        if: ${{ matrix.os != 'macos' && env.SKIP32BIT == 'false' }} # can't run 32 bit tests on OSX.
```

Both will behave exactly the same.

galargh (approved):
Thank you 🙇 Let me know if you need help setting up Tech Preview for any of the repos.

PR #413, opened 2022-11-08
chore: replace github-api-action-library with inline js

galargh:
I want to deprecate https://github.com/protocol/github-api-action-library. To make it possible, in this PR I replace usage of `protocol/github-api-action-library/find-content-by-query@v1` + some bash with inline JS script. If we end up using it more, we could move the script outside the YAML.

I tested the new version of this workflow in https://github.com/protocol/.github/actions/workflows/add-label-by-query.yml

### DIFF
--- a/.github/workflows/add-label-by-query.yml
+++ b/.github/workflows/add-label-by-query.yml
@@ -7,8 +7,12 @@ on:
         description: Query that finds issues or pull requests to which the label should be added
         required: true
       label:
-        description: Label that should be added to issues or pull requests matching the query (leave blank for dry run)
+        description: Label that should be added to issues or pull requests matching the query
+        required: true
+      dry-run:
+        description: If set to true, the workflow will only print expected actions
         required: false
+        default: 'true'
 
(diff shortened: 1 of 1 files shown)


laurentsenta (approved):
LGTM,

there was an error in the workflow

https://github.com/protocol/.github/actions/runs/3418704493

```
Couldn't add bug label to https://api.github.com/repos/pl-strflt/github-mgmt/issues/75, got: HttpError: Resource not accessible by integration
```

I guess this is expected.

galargh:
> LGTM,
> 
> there was an error in the workflow
> 
> https://github.com/protocol/.github/actions/runs/3418704493
> 
> ```
> Couldn't add bug label to https://api.github.com/repos/pl-strflt/github-mgmt/issues/75, got: HttpError: Resource not accessible by integration
> ```
> 
> I guess this is expected.

Yes! Sorry I didn't mention that. That was expected. I was trying to add a label in a repo to which `web3-bot` has no access.

PR #437, opened 2023-01-05
use GHA template expansion only in env/arg in release-check workflow

galargh:
###### Description

This PR modified `release-check.yml`. It changes it in the following ways:
- do not use global `GITHUB_ENV`
- do use steps' `GITHUB_OUTPUT`s
- do not use template expansion (`${{}}`) inside `run:` blocks
- do make other steps' outputs available to a step either through `env:` block or through `with:` block

As a result, we end up with a workflow where each step is explicitly given context. It is also given only what it requires. Finally, it is less prone to script/env var injection (we're taking in user input through `version.json@version`).

###### Testing
- [x] https://github.com/galargh/go-bitswap/actions/runs/3894978601/jobs/6649871670 + https://github.com/galargh/go-bitswap/pull/1#issuecomment-1379221832

### DIFF
--- a/.github/workflows/release-check.yml
+++ b/.github/workflows/release-check.yml
@@ -74,7 +74,7 @@ jobs:
         name: run git diff on go.mod file(s)
         if: steps.tag.outputs.exists == 'false' && steps.prev.outputs.version != ''
         env:
-          PREV_VERSION: steps.prev.outputs.version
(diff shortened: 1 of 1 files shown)


laurentsenta (approved):
Looks all goog, congrats! 👍

PR #451, opened 2023-01-30
Set source reference to `next` for some IPNI projects

masih:
The IPNI projects depending on pebble are unable to support 32-bit tests. Move those repos to `next` source reference since it has support for skipping 32-bit tests implemented.

See:
 - https://github.com/protocol/.github/issues/388

### DIFF
--- a/configs/go.json
+++ b/configs/go.json
@@ -341,10 +341,12 @@
       "target": "ipni/caskadht"
     },
     {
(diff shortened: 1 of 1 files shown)


galargh (approved):
FYI, we're planning to release next with Go 1.20 in the week of Feb 6.

PR #454, opened 2023-02-01
feat: try using custom github token for js release step

### LINKED ISSUE #430: Support protected branches for the js release flow
The js release flow makes a commit to the head of the default branch that can contain updated version numbers, monorepo sibling dependency updates, etc.

If protected branches are turned on this breaks the flow, the CI user needs to be exempted from the protection rules.

This may mean creating a custom action secret, adding it to each repo and updating the `js-test-and-release.yml` file to use this secret instead of the default `GITHUB_TOKEN` and to configure git to use a different user.

Refs: https://stackoverflow.com/questions/69263843/how-to-push-to-protected-main-branches-in-a-github-action

galargh:
Resolves #430 

With this change, the workflow will `UCI_GITHUB_TOKEN` secret for the `release` step if it's configured. Otherwise, it'll fall back to the default `GITHUB_TOKEN` the workflows have access to.

This is the quickest way I could think of to address it. One alternative would be to create a PR if a push to the default branch is prohibited.

Let me know what you think.


achingbrain:
Seems fair enough - can the `UCI_GITHUB_TOKEN` be rolled out automatically or will we need to go through every repo and update the config?

galargh:
> Seems fair enough - can the `UCI_GITHUB_TOKEN` be rolled out automatically or will we need to go through every repo and update the config?

That would be on per-repo basis.

PR #455, opened 2023-02-01
feat: allow skipping go-test on certain OSes

### LINKED ISSUE #349: Should we support configuring the OSes go-test runs on?
So far, I've encountered one instance where this would be a useful feature - see https://github.com/libp2p/go-openssl/pull/25/files#r909792651

galargh:
Resolves #349

#### Description

This PR allows configuring uCI to skip go-test on certain OSes by providing a list of OSes that should be skipped in `go-test-config.json` under `skipOSes` key.

If there's a match, the most time consuming steps of the workflow will be skipped. The job itself won't because that would require us to split the job in two, where one would be used only for figuring out whether the other should be skipped, and we don't want to introduce the overhead of starting up two jobs. 

#### Testing

- [x] tested on this commit - https://github.com/protocol/.github-test-target/commit/6206a6fa8bab0f616f83d53713e417dfe4fbc9e9

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,13 @@ To completely disable running 32-bit tests set `skip32bit` to `true` in `.github
 }
 ```
 
+If your project cannot be built on one of the supported operating systems, you can disable it by setting `skipOSes` to a list of operating systems in `.github/workflows/go-test-config.json`:
+```json
+{
+  "skipOSes": ["windows", "macos"]
+}
+```
+
 ## Technical Preview
 
 You can opt-in to receive early updates from the `next` branch in-between official Unified CI releases.
--- a/templates/.github/workflows/go-test.yml
+++ b/templates/.github/workflows/go-test.yml
@@ -47,7 +47,7 @@ jobs:
       - name: Run tests (32 bit)
         # can't run 32 bit tests on OSX.
         if: matrix.os != 'macos' &&
-          fromJSON(steps.config.outputs.json).skip32bit != true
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -56,6 +56,13 @@ To completely disable running 32-bit tests set `skip32bit` to `true` in `.github
 }
 ```
 
+If your project cannot be built on one of the supported operating systems, you can disable it by setting `skipOSes` to a list of operating systems in `.github/workflows/go-test-config.json`:
+```json
+{
+  "skipOSes": ["windows", "macos"]
+}
```
laurentsenta, line 63: Do we need `skip32bit` also?

templates/.github/workflows/go-test.yml
```diff
@@ -10,13 +10,14 @@ jobs:
...
-      SKIP32BIT: false
     runs-on: ${{ format('{0}-latest', matrix.os) }}
     name: ${{ matrix.os }} (go ${{ matrix.go }})
     steps:
       - uses: actions/checkout@v3
         with:
           submodules: recursive
+      - id: config
+        uses: protocol/.github/.github/actions/read-config@master
       - uses: actions/setup-go@v3
```
laurentsenta, line 21: Could we reduce duplication? Around here we'd set `env.SKIPPED=contains(fromJSON(steps.config.outputs.json).skipOSes, matrix.os)`. Then reuse SKIPPED below.

laurentsenta (approved):
two notes on the skip32bit and code duplication, the rest lgtm 👍

PR #458, opened 2023-02-03
chore: udpate actions and go modules

galargh:
Updates GitHub Actions actions and Go modules versions ahead of the uCI release including Go 1.20.


laurentsenta (approved):
👍

PR #489, opened 2023-03-09
feat: run go test/check on PR or push to default/protected branch

galargh:
This PR modifies the uCI templates in such a way that we don't duplicate as much work. In particular, `go-test` and `go-check` will keep running on all the pull requests BUT they will now run only on pushes to protected branches rather than on all the pushes.

###### Testing

- [x] https://github.com/protocol/.github-test-target/pull/55

### DIFF
--- a/.github/workflows/copy-workflow.yml
+++ b/.github/workflows/copy-workflow.yml
@@ -40,11 +40,16 @@ jobs:
       with:
         path: ${{ env.TEMPLATE_REPO_DIR }}
         ref: ${{ matrix.cfg.source_ref || github.ref }}
-    - name: determine GitHub default branch
+    - id: github
+      name: determine GitHub context
       working-directory: ${{ env.TARGET_REPO_DIR }}
+      env:
+        REPO: ${{ matrix.cfg.target}}
       run: |
-        default_branch=$(git remote show origin | awk '/HEAD branch/ {print $NF}')
-        echo "DEFAULT_BRANCH=$default_branch" >> $GITHUB_ENV
+        default_branch="$(git remote show origin | awk '/HEAD branch/ {print $NF}')"
+        protected_branches="$(gh api -X GET "/repos/$REPO/branches" -f protected=true | jq -c 'map(.name)')"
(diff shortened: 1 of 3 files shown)


.github/workflows/copy-workflow.yml
```diff
@@ -105,11 +116,9 @@ jobs:
         CONTEXT: |
           {
             "config": ${{ toJSON(matrix.cfg) }},
-            "github": {
-              "default_branch": "${{ env.DEFAULT_BRANCH }}"
-            }
+            "github": ${{ toJSON(steps.github.outputs.json) }}
```
galargh, line 119:
```suggestion
            "github": ${{ toJSON(fromJSON(steps.github.outputs.json)) }}
```

.github/workflows/copy-workflow.yml
```diff
@@ -105,11 +120,9 @@ jobs:
         CONTEXT: |
           {
             "config": ${{ toJSON(matrix.cfg) }},
-            "github": {
-              "default_branch": "${{ env.DEFAULT_BRANCH }}"
-            }
+            "github": ${{ toJSON(fromJSON(steps.github.outputs.json)) }}
           }
-        TEMPLATE_ENGINE: s#\$\{\{\{\s*(.*?)\s*\}\}\}#`jq -cjn 'env.CONTEXT | fromjson.$1'`#ge
+        TEMPLATE_ENGINE: s#\$\{\{\{\s*\.?(.*?)\s*\}\}\}#`jq -cjn 'env.CONTEXT | fromjson | .$1'`#ge
```
laurentsenta, line 125: we're making the `.` optional then normalize the output so that it always starts with a `.`, correct?

laurentsenta (approved):

.github/workflows/copy-workflow.yml
galargh: Yeah. I felt that when you end up using full blown jq in the template, it looked weird without the leading dot. It's purely about aesthetics.

PR #491, opened 2023-03-09
feat: do not hardcode Go version in workflows

galargh:
This attempts to read required Go version from go.mod (using `go mod edit -json`) rather than hardcoding it in the workflow.

###### Testing

- [x] Successfully deployed to .github-test-target - https://github.com/protocol/.github-test-target/pull/57


.github/workflows/release-check.yml
```diff
@@ -14,9 +14,12 @@ jobs:
       - run: echo "EOF=EOF$RANDOM" >> $GITHUB_ENV
         shell: bash
       - uses: actions/checkout@v3
+      - id: go-mod
+        # TODO: change ref to next before merging
+        uses: protocol/.github/.github/actions/read-go-mod@no-go
```
galargh, line 19:
```suggestion
        uses: protocol/.github/.github/actions/read-go-mod@next
```

templates/.github/workflows/go-test.yml
```diff
@@ -14,34 +14,52 @@ jobs:
...
+        #   subsequent 'shell: bash' steps will use msys2 instead of gitbash
+        run: echo "C:/msys64/usr/bin" >> $GITHUB_PATH
       - uses: actions/checkout@v3
         with:
           submodules: recursive
       - id: config
         uses: protocol/.github/.github/actions/read-config@master
+      - id: go-mod
+        # TODO: change ref to next before merging
+        uses: protocol/.github/.github/actions/read-go-mod@no-go
```
galargh, line 40:
```suggestion
        uses: protocol/.github/.github/actions/read-go-mod@next
```

templates/.github/workflows/go-test.yml
```diff
@@ -14,34 +14,52 @@ jobs:
...
+          if [[ "$MATRIX_GO" == "this" ]]; then
+            echo "version=$GO_MOD_VERSION.x" >> $GITHUB_OUTPUT
+          elif [[ "$MATRIX_GO" == "next" ]]; then
+            MAJOR="${GO_MOD_VERSION%.[0-9]*}"
+            MINOR="${GO_MOD_VERSION#[0-9]*.}"
+            echo "version=$MAJOR.$(($MINOR+1)).x" >> $GITHUB_OUTPUT
+          else
+            echo "version=$MATRIX_GO" >> $GITHUB_OUTPUT
+          fi
+        shell: bash
```
laurentsenta, line 55: Are we constraining the version somewhere else? I guess that means a maintainer can break their unified CI by using the latest go version?

templates/.github/workflows/go-test.yml
```diff
@@ -14,34 +14,52 @@ jobs:
...
       GORACEFLAGS: -shuffle=on
     runs-on: ${{ fromJSON(vars[format('UCI_GO_TEST_RUNNER_{0}', matrix.os)] || format('"{0}-latest"', matrix.os)) }}
     name: ${{ matrix.os }} (go ${{ matrix.go }})
     steps:
+      - name: Use msys2 on windows
+        if: matrix.os == 'windows'
+        shell: bash
+        # The executable for msys2 is also called bash.cmd
+        #   https://github.com/actions/virtual-environments/blob/main/images/win/Windows2019-Readme.md#shells
+        # If we prepend its location to the PATH
```
laurentsenta, line 30: til github will turn that `>>` call into a prepend 👍

laurentsenta (approved):
A question on version constraints that you probably already thought about, rest looks great 👍

templates/.github/workflows/go-test.yml
galargh: This is fixed now. The versions to use can be configured with arguments to the reusable go-test workflow. One can use `this`, `next`, `prev` keywords or specify the exact versions to use. Of course, the dispatched go-test workflow cannot be modified now but one of the features of Unified CI 2.0 will enable it (make dispatched workflows modifiable - to be exact).

PR #494, opened 2023-03-14
feat: add go test summaries

galargh:
I propose that we add MD test reports to the `go-test` workflow. They will appear in the `Summary` section of the workflow runs. The summaries will include cumulative information about test runs as well as details of failures/errors encountered during the runs. 

Here's an example summary: https://github.com/protocol/.github-test-target/actions/runs/4415546379/attempts/1#summary-11987752699

### DIFF
--- a/templates/.github/workflows/go-test.yml
+++ b/templates/.github/workflows/go-test.yml
@@ -46,7 +46,7 @@ jobs:
         env:
           GOFLAGS: ${{ format('{0} {1}', env.GOTESTFLAGS, env.GOFLAGS) }}
         with:
-          run: gotestsum --jsonfile=gotest.All.json --packages=./... -- -v
+          run: gotestsum --jsonfile=gotest.All.json --format=standard-verbose --packages=./...
       - name: Run tests (32 bit)
         # can't run 32 bit tests on OSX.
         if: matrix.os != 'macos' &&
@@ -59,7 +59,7 @@ jobs:
         with:
           run: |
             export "PATH=$PATH_386:$PATH"
-            gotestsum --jsonfile=gotest.32-bit.json --packages=./... -- -v
+            gotestsum --jsonfile=gotest.32-bit.json --format=standard-verbose --packages=./...
       - name: Run tests with race detector
         # speed things up. Windows and OSX VMs are slow
         if: matrix.os == 'ubuntu' &&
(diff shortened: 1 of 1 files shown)


marten-seemann (commented):
A few thoughts:

1. That's quite a lot of additional code.
2. There's value in using the same thing that people use locally (i.e. `go test`).
3. How confident are we in `gotestsum` not introducing new problems?
4. The output of the `gotestsum` command looks very different from the `go test` output. Not sure how it would look for failures.
5. I'm not convinced that the summary adds a lot of value. It counts the number of tests and failures, but if you want to see why a test failed, you'll still have to go back to the test output.

galargh:
1. Yes. Unfortunately, the number of different configurations we run the tests adds complexity. 
2. We use `gotestsum` only as a wrapper so that we can get `JSON` report while preserving the standard stdout. If we didn't care about `stdout`, we'd be fine with `go test -v -json | tee gotest.json` or something along those lines.
3. We've been using `gotestsum` in `ipfs/kubo` since 2019 - https://github.com/ipfs/kubo/commit/93c75321fbb03c1ae427343c4bcda5718fcbfda4.
4. Sorry about that. I thought the default was maintaining the standard stdout. I requested it specifically in https://github.com/protocol/.github/pull/494/commits/a7b74645bc0829603ba95728e59f74b6c69191fe. Here's an example run with the updated settings: https://github.com/protocol/.github-test-target/actions/runs/4424453098/jobs/7758299695
5. The summary would include details about failures/errors. Here's an example from a test run which has a failure: https://github.com/ipfs/kubo/actions/runs/4407775163/attempts/1#summary-11965710632 (it's from Kubo so there's only one test run we report on but the format is the same).

Another consideration is that it takes a minute to create all the reports - https://github.com/protocol/.github-test-target/actions/runs/4424453098/jobs/7758406237 (this is a repo with 2 Go modules).

marten-seemann:
> 1. Yes. Unfortunately, the number of different configurations we run the tests adds complexity. That's also why it takes a minute - https://github.com/protocol/.github-test-target/actions/runs/4424453098/jobs/7758406237 (this is a repo with 2 Go modules).

I hadn't seen that. That seems like a very long time for doing some JSON parsing. I wouldn't have expected this to take more than a few ms (plus setup time of the runner of course).

> 5\. The summary would include details about failures/errors. Here's an example from a test run which has a failure: https://github.com/ipfs/kubo/actions/runs/4407775163/attempts/1#summary-11965710632 (it's from Kubo so there's only one test run we report on but the format is the same).

Does it also deal with timeouts and panics during a test?

PR #496, opened 2023-03-15
ci: try to perform fewer pushes to web3-bot/sync branches

galargh:
I propose to update the copy workflow as follows to reduce the number of pushes we do to unmerged web3-bot/sync branches.

###### Before

1. Check out the default branch.
2. Update Unified CI files.
3. If any files are different than in the default branch:
    1. Force check out `web3-bot/sync`.
    2. Force push `web3-bot/sync`.

In this setup, we push `web3-bot/sync`s to targets every time we merge something to uCI master as long as the uCI files in default branch of targets are not up-to-date.

###### After

1. Check out the default branch.
2. If `web3-bot/sync` doesn't exist, create it from the default branch.
3. Check out `web3-bot/sync`.
4. If `web3-bot/sync` is behind the default branch:
    1. Try rebasting the `web3-bot/sync` branch onto the default branch.
    2. If the merge fails, force reset `web3-bot/sync` to the default branch.
5. Update Unified CI files. 
6. If any files were updated in the previous step, force push `web3-bot/sync`.

###### Other considerations

In this setup, we push `web3-bot/sync`s to targets only when uCI files change in that branch.

There is one scenario where the updated workflow would require manual attention, whereas previously, it didn't. The scenario is as follows:
1. The default branch of a target is broken
2. We push `web3-bot/sync` update
3. CI fails and `web3-bot/sync` is not automatically merged
4. The default branch is fixed
5. We modify something in uCI master
6. Previously, that would initiate a force push to `web3-bot/sync` and subsequent automatic merge if the CI passed. After the updates, one would have to manually merge default branch into `web3-bot/sync`. 

I think it's a small price to pay for fewer pushes. Especially since in this scenario, manual attention is needed already anyway since the default branch is broken to begin with.

###### Testing

I want first to learn your thoughts about this proposal. I'm going to test the updated workflow before deploying it.


marten-seemann:
> 1. Try merging the default branch into `web3-bot/sync`.

I consider this an antipattern. Merging the default branch leads to very hard to understand commit histories.

galargh:
> > 1. Try merging the default branch into `web3-bot/sync`.
> 
> I consider this an antipattern. Merging the default branch leads to very hard to understand commit histories.

How about `fast-forward` only?

galargh:
> > 1. Try merging the default branch into `web3-bot/sync`.
> 
> I consider this an antipattern. Merging the default branch leads to very hard to understand commit histories.

I see. How would you feel about `git rebase $DEFAULT_BRANCH` with a fallback to `git reset $DEFAULT_BRANCH --hard`? This would mean we're still force pushing whenever there are differences in uCI files but it would reduce the number of those force pushes because we would only perform them if uCI files changed, not when the target default branch moved.

galargh:
@marten-seemann A friendly reminder about the proposal in https://github.com/protocol/.github/pull/496#issuecomment-1469863960. It's not urgent.

galargh:
I'm going to merge this into my Work In Progress Unified CI 2.0 branch so that I can experiment and test it. If you have any reservations for `git rebase || git reset --hard` approach, please let me know :)

PR #509, opened 2023-03-31
fix: semver comparison for prereleases in release-check

### LINKED ISSUE #508: Release Check workflow not comparing the correct versions for Golang RCs
For example, https://github.com/ipfs/boxo/pull/234 is comparing v0.8.0-rc3 to v0.7.0 rather than showing me the delta against v0.8.0-rc2

galargh:
Closes #508

With this change, the release-check will take into account prerelease versions when looking for a version to compare to when releasing a prerelease version.

The behaviour for releasing non-prerelease versions does not change.


marten-seemann:
Comparing to the last actual release (i.e. the current behavior) is the correct thing to do.
There's no API guarantees for pre-release versions. Comparing to pre-releases will make us miss breaking changes.

galargh:
I'm only proposing to compare pre-releases to pre-releases where I think it could make sense. I'll let @aschmahmann to advocate for it, given https://github.com/protocol/.github/issues/508.

aschmahmann:
> There's no API guarantees for pre-release versions. Comparing to pre-releases will make us miss breaking changes.

How? When you go from v0.X.0 to v0.X+1.0-rc1 you'll have a bunch of breaking changes. Then at some point you make an v0.X.+1.0-rc2, don't you want to know the delta vs rc1? If rc1 had a lot of breaking changes then rc2,3,... all have a ton of noise in the checks.

v0.X.0-rc1 would still show changes from x0.X.0 and v0.X.1 (or v0.X+1.0) would also show the changes from v0.X.0.

marten-seemann:
> I'm only proposing to compare pre-releases to pre-releases where I think it could make sense. I'll let @aschmahmann to advocate for it, given #508.

As long as we're still showing the full list of changes since v0.X-1 when cutting v0.X that makes sense to me.

PR #513, opened 2023-04-06
remove verbose mode in go-test

Jorropo:
This is extremely spamy and not usefull.
Also trying to open the CI results on boxo crashes by browser or takes a huge while due to a way too big test suite.

I think the default mode is pretty readable and low noise (only output what we need).


galargh:
It'd be nice if we could make it configurable. We should also target the `next` branch with a change like that. I can look into it after IPFS Thing.

galargh:
Thanks for the reminder, I'm going to close this now in favor of https://github.com/protocol/.github/pull/520

PR #520, opened 2023-05-12
feat: make -v and -shuffle configurable options

galargh:
This is an alternative to https://github.com/protocol/.github/pull/513 which keeps the defaults as they were while allowing to opt-out of -v and -shuffle=on flags.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,14 @@ If your project cannot be built on one of the supported operating systems, you c
 }
 ```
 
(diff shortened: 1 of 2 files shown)


Jorropo (commented):
SGTM

Jorropo:
@galargh is this blocked on anything ? thx

galargh:
Just forgot, rolling out to boxo now.

PR #521, opened 2023-05-15
add quic-go/perf to Unified CI

marten-seemann:


marten-seemann:
The quic-go org requires the Google CLA to be signed. We have a CLA bot that checks for that on every PR.
@protocol/ipdx, would you be able to sign that CLA on behalf of the web3-bot?

galargh:
Oh, sorry, I've been meaning to get back to you on this one. This one's a little tricky since `web3-bot` GitHub user is associated with a Google group rather than a Google account. I guess one way to go around that would be to create a concrete account for the bot; we could possibly still use the group email as the primary email for the account in GitHub. Do you know if we had to solve it for some other bot somewhere around PL already? 

We could also file a bug with Google claiming that our [external robot](https://opensource.google/documentation/reference/cla#robots) only ever touches CI files, but I'm not sure if that's going to work looking at the examples they've given.

galargh:
Moving this to https://github.com/pl-strflt/uci/issues/33

PR #536, opened 2023-07-07
Add Singularity

willscott:
singularity is used as a client pushing data into filecoin.

cc @masih  @xinaxu

### DIFF
--- a/configs/go.json
+++ b/configs/go.json
@@ -19,6 +19,9 @@
   },
   "repositories": [
     {
+      "target": "data-preservation-programs/singularity"
+    },
+    {
       "target": "filecoin-project/docker-hub-exporter"
     },
     {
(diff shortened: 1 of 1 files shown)


configs/go.json
```diff
@@ -122,6 +122,9 @@
       ],
       "deploy_versioning": false
     },
+    {
+      "target": "data-preservation-programs/singularity"
```
masih, line 126: @xinaxu  for this to function @web3-bot needs to be given write permissions to the repo.

masih (approved):

configs/go.json
xinaxu: Invite sent to @web3-bot

galargh: Thank you, and sorry for the delay. Here's the uCI PR - https://github.com/data-preservation-programs/singularity/pull/78

